PR #3, opened 2026-08-18
Fix local pairing commands using the Vite API target

### LINKED ISSUE #1: Local macOS setup issues: OAuth return URL rejected and generated computer pairing command targets production
## Summary

While setting up **Cumora locally on macOS**, I encountered two issues that prevented **GitHub login** and **BYOA computer pairing** from working with the default local development setup.

The web app and API started successfully with:

```bash
npm run dev:all
```

Local environment:

* Web: `http://localhost:5180`
* API: `http://localhost:5181`
* OS: macOS
* BYOA engine: Codex

---

## Issue 1: GitHub OAuth Returns HTTP 400

GitHub OAuth HTTP 400

After configuring `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET` in the root `.env`, clicking **GitHub Login** resulted in:

```text
GET /auth/start/github?return=http%3A%2F%2Flocalhost%3A5180%2F → 400
```

The OAuth credentials were loaded correctly.

The request was rejected because the default/example `CUMORA_AUTH_RETURN_ALLOWLIST` contains:

```text
http://localhost:5173/
```

However, the Vite development server started by `npm run dev:all` actually runs on:

```text
http://localhost:5180/
```

### Workaround

Adding the actual Vite development origin fixed the problem:

```env
CUMORA_AUTH_RETURN_ALLOWLIST=http://localhost:5180/
```

For example:

```env
CUMORA_AUTH_RETURN_ALLOWLIST=https://app.cumora.ai/,https://admin.cumora.ai/,http://localhost:5173/,http://localhost:5180/,http://127.0.0.1:47823/auth/done,cumora://auth
```

The GitHub OAuth App callback URL also needs to be configured as:

```text
http://localhost:5181/api/auth/callback/github
```

### Expected Behavior

The checked-in local development defaults and `.env.example` should match the actual Vite port used by:

```bash
npm run dev:all
```

### Suggested Improvements

* Add `http://localhost:5180/` to the example development OAuth return allowlist.

* Remove or explain the older `http://localhost:5173/` entry if it is still required.

* Explicitly document the required GitHub OAuth callback URL:

  ```text
  http://localhost:5181/api/auth/callback/github
  ```

* Consider automatically allowing the configured Vite development origin when `NODE_ENV=development`, while keeping strict allowlisting in production.

* Include the rejected `return` URL in development-mode error logs to make this issue easier to diagnose.

---

## Issue 2: Generated Pairing Command Sends a Local Token to Production

Invalid pairing token

After choosing **Add a computer**, the local UI generated a command similar to:

```bash
npx cumora@latest agent computer \
  --pair <token> \
  --engine codex \
  --install-service
```

Running it returned:

```text
cumora: POST /api/computers/pair → HTTP 400 {"error":"invalid pairing token"}
```

The pairing token itself was valid and existed in the local database.

The problem is that the Cumora CLI defaults to:

```text
https://api.cumora.ai
```

Because the generated command did not contain a `--server` argument, the pairing token generated by the **local Cumora instance** was sent to the **production API** instead.

### Workaround

Explicitly passing the local API server fixes the issue:

```bash
npx cumora@latest agent computer \
  --pair <token> \
  --server http://localhost:5181 \
  --engine codex \
  --install-service
```

If pairing another computer on the same LAN, the local server's reachable LAN address must be used instead of `localhost`.

For example:

```bash
npx cumora@latest agent computer \
  --pair <token> \
  --server http://192.168.x.x:5181 \
  --engine codex \
  --install-service
```

### Expected Behavior

A pairing command generated by a locally running Cumora UI should target the **same backend instance that generated the pairing token**.

### Likely Cause

In local Vite development, `getServerOrigin()` returns an empty string because browser API requests use relative URLs through the Vite proxy.

This is valid for browser requests:

```text
Browser
  ↓
/api/...
  ↓
Vite Proxy
  ↓
http://localhost:5181
```

However, the command-generation UI only appends `--server` when `getServerOrigin()` returns a non-empty value.

As a result, the generated CLI command becomes:

```bash
npx cumora@latest agent computer --pair <token> ...
```

The CLI then falls back to its default backend:

```text
https://api.cumora.ai
```

So the actual flow becomes:

```text
Local UI
  ↓
Generate local pairing token
  ↓
Generated CLI command has no --server
  ↓
CLI defaults to https://api.cumora.ai
  ↓
Local token is sent to production
  ↓
invalid pairing token
```

An empty browser API origin is therefore valid for browser requests, but it is **not sufficient for a separately launched CLI process**.

### Suggested Improvements

When generating CLI pairing commands, resolve an explicit API origin.

Suggested behavior:

1. If `getServerOrigin()` is explicitly configured, use it.

2. In Vite development, use `CUMORA_DEV_API_TARGET` or the known local API origin, for example:

   ```text
   http://localhost:5181
   ```

3. For same-origin production deployments, use:

   ```js
   window.location.origin
   ```

4. Make the API server address visible and editable before copying the pairing command.

5. Consider encoding the intended server origin into the pairing payload or generating a self-contained pairing URL.

6. Improve the CLI error when a pairing token is rejected by the default production server.

For example:

```text
Pairing token was not found on https://api.cumora.ai.

If this token was generated by a local or self-hosted Cumora instance,
rerun the command with:

  --server <server-url>
```

---

## Environment

```text
OS: macOS

Web:
http://localhost:5180

API:
http://localhost:5181

Start command:
npm run dev:all

BYOA engine:
Codex
```

Both issues appear to be caused by differences between the **local development defaults** and the ports/origins actually used by the current Vite development setup.

ssynb:
## Summary

Fixes #1.

When the web app runs with `npm run dev:all`, the renderer uses Vite's relative `/api` proxy, so `getServerOrigin()` is intentionally empty. Pairing and repair commands copied from the UI therefore omitted `--server`, causing the CLI to fall back to `https://api.cumora.ai` instead of the local API.

This change:

- exposes the configured Vite API proxy target to the renderer;
- uses that target when generating computer pairing/repair commands in onboarding and computer settings;
- adds the actual `:5180` renderer origin to the `.env.example` OAuth return allowlist.

The normal browser API and WebSocket paths remain relative in Vite development.

## Validation

- `npm run typecheck` ✅
- `npm run lint` ✅ (existing Biome configuration deprecation notice only)
- `npm run server:typecheck` ✅
- `npm run build` ✅
- `npm run guard:big-brain` ✅
- `npm test` ⚠️ the clean environment has no `OPENAI_API_KEY`/local service dependencies; unrelated server tests fail during startup
- `npm run guard:llm-tracked` ⚠️ reports pre-existing server call sites outside its allowlist; this PR does not touch them

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -125,7 +125,7 @@ R2_URL_SIGNING_SECRET=
 # without this we'd be open-redirect bait. Comma-separated; prefix match.
 # Must include every host the SPA may be served at (app + admin) plus
 # the Electron loopback page and deep-link.
-# CUMORA_AUTH_RETURN_ALLOWLIST=https://app.cumora.ai/,https://admin.cumora.ai/,http://localhost:5173/,http://127.0.0.1:47823/auth/done,cumora://auth
+# CUMORA_AUTH_RETURN_ALLOWLIST=https://app.cumora.ai/,https://admin.cumora.ai/,http://localhost:5180/,http://localhost:5173/,http://127.0.0.1:47823/auth/done,cumora://auth
 #
 # Public base used to construct invitation accept URLs
 # (`<base>/invite/<token>`). Defaults to CUMORA_AUTH_DONE_URL — set this
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -93,11 +93,42 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
     timeout-minutes: 10
-    # env.ts hard-requires OPENAI_API_KEY (process.exit(1) otherwise), and
-    # unit tests boot it. Provide a dummy so the boot passes — matches the
-    # value build.yml's quality job uses. No real key; unit tests never
-    # call a provider.
+    # `npm test` (server/src/__tests__ + workers) boots env.ts and opens
+    # Postgres/Redis connections, so this job needs the same service
+    # containers + env as build.yml's quality job — without them tests
+    # hang on ECONNREFUSED until the job times out. Dummy secrets only;
+    # unit tests never reach a real provider. (pgvector isn't needed here —
+    # build.yml runs `npm test` before enabling the extension.)
+    services:
+      postgres:
+        image: pgvector/pgvector:pg16
+        env:
+          POSTGRES_USER: cumora
+          POSTGRES_PASSWORD: cumora
+          POSTGRES_DB: cumora_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U cumora -d cumora_test"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+      redis:
+        image: redis:7-alpine
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
     env:
+      DATABASE_URL: postgres://cumora:cumora@localhost:5432/cumora_test
+      REDIS_URL: redis://localhost:6379
+      RESEND_API_KEY: ''
+      EMAIL_DOMAIN: cumora.local
+      EMAIL_INBOUND_HMAC_SECRET: ci-test-secret
+      AUTH_SECRET: ci-test-auth-secret
       OPENAI_API_KEY: ci-test-openai-key
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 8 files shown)


yetone:
The failing `Unit tests` check was a CI config gap on our side, not an issue with this PR: the PR workflow ran the worker tests (`workers/**/*.test.ts`) without installing the email-gate worker's own dependency (`postal-mime`). Fixed on `main` (2873e50). Re-triggering CI here so it picks up the fixed workflow.

PR #23, opened 2026-08-19
feat(computer): add Grok Build as a BYOA engine

storyicon:
Adds Grok Build (`grok`) as a third pairable BYOA engine next to Claude Code and Codex.

- Detect `grok` on PATH or `~/.grok/bin`
- Persistent session: `grok agent --always-approve --no-leader stdio` (ACP)
- One-shot fallback: `grok -p` when `CUMORA_GROK_ARGS` / `CUMORA_GROK_NO_ACP=1` is set, and on Windows (same JSON-RPC-over-`.cmd` trap as Codex)
- Pairing UI, agent engine picker, `byoa-grok` ledger source, doctor probe
- Optional pin via `CUMORA_DEFAULT_GROK_MODEL` / per-agent `--model`

Grok's ACP `session/prompt` is one-in-flight, so same-turn steer is a no-op; the daemon coalesces the ping onto the next wake.

### DIFF
--- a/agent-cli/src/cli.ts
+++ b/agent-cli/src/cli.ts
@@ -21,7 +21,7 @@ async function main(): Promise<void> {
     'Usage:\n' +
     '  npx cumora@latest agent computer --pair <code> [--server <url>]   pair this machine\n' +
     '  npx cumora@latest agent computer [--server <url>]                 start the daemon\n\n' +
-    'Needs `claude` (Claude Code) or `codex` on PATH. Get a pairing code from\n' +
+    'Needs `claude` (Claude Code), `codex`, or `grok` (Grok Build) on PATH. Get a pairing code from\n' +
     'Cumora → You → Computers → Add a computer.\n',
   )
   process.exit(argv.length ? 1 : 0)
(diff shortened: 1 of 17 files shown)


terry-li-hm:
Tested the adapter on macOS 26 (Darwin 25.5), Node from the repo's own toolchain, `grok` 1.0.5 at `~/.grok/bin/grok`, OAuth login, under a clean `env -i HOME USER SHELL TERM PATH` like the daemon uses. Drove `engine.ts` directly with a small tsx harness rather than through a paired computer, so this covers detection, doctor, one-shot and the ACP session, not the server pairing UI.

- `detectEngines()` → `["claude","codex","grok"]` (resolved via `~/.grok/bin`, not on PATH)
- `runEngineDoctor`: big 16.0s OK, small 6.2s OK, wake 2.0s OK
- `run()` one-shot (`streaming-messages-json`): exit 0, session id parsed, usage `{input 21065, output 40, cache_read 5760}`, `model: grok-4.6`, `apiKeySource: oauth`
- `startSession()` over `grok agent --always-approve --no-leader stdio`: standing prompt carried (`carriesStandingPrompt: true`), two turns on one session id, usage reported per turn (`model` comes back null on the ACP path, so the ledger would price on the pin rather than the real model; worth noting)

No failures. Happy to re-test through a paired computer if you want a second data point before merging.

PR #55, opened 2026-08-21
feat(llm): add Novita provider via a Responses-API translation adapter

jax-novita:
## Summary

Adds Novita as an LLM provider for Cloud agents. Novita's API is OpenAI-compatible only at the Chat Completions layer, not the Responses API that Cumora's agent turn/convene/triage call sites are written against, so a base-URL swap alone would 404.

## Changes

- `server/src/novita.ts`: translates Responses-shaped requests (instructions/input/tools/streaming) into Chat Completions calls, and translates the result (including streamed `ChatCompletionChunks`) back into the exact `ResponseStreamEvent` sequence `turn-stream.ts` already consumes, so no existing call site changes.
- `server/src/llm.ts`: `getLlmClient()` routes per-call based on a `novita/<model>` prefix on the model id (`withNovitaRouting`), so an agent opts in by setting that as its model.
- `server/src/env.ts`: adds the Novita API key/base URL env vars, soft-disabled like the other optional integrations when unset.
- `src/components/AgentEditor.tsx`: documents the `novita/<model>` convention in the model field hint for Cloud agents.
- `server/src/__tests__/novita.test.ts`: adapter tests covering request/response translation and streaming.
- Fixes a bug found during testing: the OpenAI SDK's `buildRequest()` guards timeout validation with `'timeout' in options` rather than a nullish check, so always including the key (even as `undefined`) threw `OpenAIError: timeout must be an integer` before any request was sent, surfacing upstream as a misleading connection/timeout failure. Only `signal`/`maxRetries`/`timeout` are now set on the per-call options object when a real value is present.

## Verification

- `server/src/__tests__/novita.test.ts`: 6 focused adapter tests, all passing.
- Live request through the adapter against Novita's API succeeded end-to-end (model `deepseek/deepseek-v4-pro-0813`, non-streaming completion).

### DIFF
--- a/server/src/__tests__/novita.test.ts
+++ b/server/src/__tests__/novita.test.ts
@@ -1,0 +1,223 @@
+/**
+ * Unit tests for the Novita Chat-Completions ↔ Responses-API translation
+ * adapter (server/src/novita.ts). No network access: `chat.completions.create`
+ * is stubbed via `__setNovitaClientOverrideForTesting` so these tests assert
+ * only the translation logic, not live Novita behavior (that's QA's job with
+ * a real NOVITA_API_KEY).
+ *
+ * Run: node --import tsx --test server/src/__tests__/novita.test.ts
+ */
+
+import assert from 'node:assert/strict'
+import { test } from 'node:test'
+import type OpenAI from 'openai'
+import type { ChatCompletionChunk } from 'openai/resources/chat/completions'
+import type { ResponseInputItem, ResponseStreamEvent } from 'openai/resources/responses/responses'
+import { applyResponseStreamEvent, consumeResponseStream, newResponseStreamState } from '../agents/turn-stream.js'
+import {
+  __setNovitaClientOverrideForTesting,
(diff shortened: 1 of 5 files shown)


yetone:
Reviewed — the translation shim is solid (nice touch testing it against the real `turn-stream.ts` consumer, and the `'timeout' in options` SDK finding is appreciated).

One behavioral gap, which I've fixed directly on your branch (333683c) since `maintainerCanModify` was on: `withNovitaRouting` routed on the model prefix alone, but `env.ts` documents that an unset `NOVITA_API_KEY` falls back to the tenant's normal client. As written, an unconfigured deployment sent the call to `api.novita.ai` with an empty bearer and the agent died on an unexplained 401 — on hosted Cumora (which doesn't set the key) that's every user who tries the `novita/` hint the PR adds to AgentEditor. The guard makes the documented fallback real, with a one-time log so operators can see why the model isn't resolving to Novita.

Merging once CI is green on that commit. Thanks for the contribution!

PR #61, opened 2026-08-23
fix(auth): trust only verified Apple emails

KIDA-MNESIA:
## Summary

- remove the client-provided email from the native Sign in with Apple request contract
- resolve returning Apple users from the persisted provider `sub` mapping
- allow first-time creation or cross-provider linking only when the signed Apple token contains a verified email
- add regression coverage for missing, unverified, verified, and returning-user email cases

## Security impact

Previously, when an unlinked Apple identity token did not contain a usable email, the native endpoint accepted a request-body fallback email. That untrusted value could reach the existing cross-provider auto-link lookup and bind the Apple identity to an account with the supplied email.

This change keeps client callback metadata out of the identity boundary. Existing Apple identities authenticate through their persisted `sub` mapping; unlinked identities require an Apple-signed `email_verified=true` claim before account creation or linking.

## Validation

Passed locally:

- `node --import tsx --test server/src/__tests__/oauth-apple-email.test.ts` (5/5)
- `npm run lint`
- `npm run typecheck`
- `npm run server:typecheck`
- `npm run build`
- `npm run guard:big-brain`
- `npm run guard:llm-tracked`
- `git diff --check`

Not completed locally:

- full `npm test` requires the repository's configured test services and environment
- `npm run test:integration` requires `INTEGRATION_DATABASE_URL`, Postgres, and Redis

The PR is left as a draft until the repository CI completes the full suite.

### DIFF
--- a/server/src/__tests__/oauth-apple-email.test.ts
+++ b/server/src/__tests__/oauth-apple-email.test.ts
@@ -1,0 +1,58 @@
+/**
+ * Regression tests for the Sign in with Apple account-linking boundary.
+ *
+ * Run: node --import tsx --test server/src/__tests__/oauth-apple-email.test.ts
+ */
+import { test } from 'node:test'
+import assert from 'node:assert/strict'
+import { resolveTrustedAppleEmail } from '../apple.js'
+
+test('unlinked Apple identity rejects an email-less token even when the client supplies a victim email', () => {
+  const attackerInput = {
+    linkedEmail: null,
+    tokenEmail: null,
+    tokenEmailVerified: false,
+    // Extra request metadata must never enter the trusted selector.
+    fallbackEmail: 'victim@example.com',
+  }
+
+  assert.throws(
+    () => resolveTrustedAppleEmail(attackerInput),
+    /verified email not available for unlinked identity/,
+  )
+})
+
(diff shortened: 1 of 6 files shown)


yetone:
Reviewed and merging — thank you for finding and fixing this properly.

Confirmed the hole is real on `main`: `handleAppleNativeSignIn` resolved the email as `claims.email ?? args.fallbackEmail ?? knownEmail`, so an Apple-signed token with no email claim and an unlinked `sub` let a request-body email reach `findOrCreateUserByProfile`, which cross-links by email. That is an account-takeover path, not just a hygiene issue.

The shape of the fix is right where it matters: a linked `sub` resolves from its persisted email, so returning users keep signing in even when Apple omits the claim (that was my first concern — it's handled), and only an unlinked identity is gated on a signed `email_verified` claim. Removing `email` from the request contract entirely, rather than just ignoring it, is the part that makes this stay fixed.

You left it as a draft pending full CI — that has since gone green (6/6, integration included), so I've marked it ready and merged rather than leave a live auth hole open over the weekend. Shipping it in the next release now.

The managed-school-account tradeoff you flagged in the comment is the right call to make explicitly.

PR #66, opened 2026-08-24
fix(agents): restrict attachment fetches to storage

KIDA-MNESIA:
## Summary

- accept chat attachments only when they resolve to the server-managed `attachments/` storage namespace
- discard client-selected retrieval URLs and persist a fresh URL minted from the validated storage key
- apply the same fail-closed key policy before agents read text files or materialize message images
- reject storage redirects and unsafe/traversal key forms
- add regression tests for private/public attacker URLs, namespace confusion, mismatched references, and path traversal

## Security impact

Previously, an authenticated conversation member could submit a text-like attachment with an arbitrary HTTP(S) URL. When an agent processed the message, the server fetched that URL and injected the response into agent context, allowing SSRF against services reachable from the server network.

After this change, only a server-generated `attachments/…` key can select attachment content. Both the write boundary and the agent-read boundary regenerate URLs from that key. Legacy or external attachment URLs remain visible as message metadata but are not downloaded by the server or agent runtime.

## Validation

Passed locally:

- `node --import tsx --test server/src/__tests__/storage-keys.test.ts` (7/7)
- `npm run lint` (one pre-existing informational finding in `server/src/agents/cli.ts`)
- `npm run typecheck`
- `npm run server:typecheck`
- `npm run build`
- `npm run guard:big-brain`
- `npm run guard:llm-tracked`
- `git diff --check`

Local environment limitations:

- full `npm test` could not complete because this checkout has no required `OPENAI_API_KEY`; the spawned test runner was stopped after recording the import-time failures
- `npm run test:integration` was skipped by the repository runner because `INTEGRATION_DATABASE_URL` is unset

Repository CI passed all six checks, including the full unit and integration suites.

### DIFF
--- a/server/src/__tests__/storage-keys.test.ts
+++ b/server/src/__tests__/storage-keys.test.ts
@@ -1,0 +1,87 @@
+/**
+ * Regression tests for message attachment storage references.
+ *
+ * Run: node --import tsx --test server/src/__tests__/storage-keys.test.ts
+ */
+import { test } from 'node:test'
+import assert from 'node:assert/strict'
+import {
+  messageAttachmentStorageKey,
+  normalizeStorageKey,
+  storageKeyFromPublicUrl,
+} from '../storage-keys.js'
+
+const PUBLIC_BASE = 'https://cdn.cumora.test/private-files'
+
+test('rejects arbitrary text-attachment URLs, including private network targets', () => {
+  for (const url of [
+    'http://127.0.0.1:5181/admin',
+    'http://169.254.169.254/latest/meta-data',
+    'http://10.0.0.8/internal',
+    'https://attacker.example/payload.txt',
+  ]) {
+    assert.equal(messageAttachmentStorageKey({ url }, PUBLIC_BASE), null)
+  }
+})
+
+test('accepts local and configured-public-base attachment URLs', () => {
+  assert.equal(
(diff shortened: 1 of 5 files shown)


yetone:
Reviewed and merging — this is a real one, thank you.

Verified the hole on `main` rather than taking the report at face value: `POST /conversations/:id/messages` accepts `attachment.url` verbatim and treats `attachment.key` as optional, so omitting the key makes `freshAttachmentUrl` fall through to the client's URL, which `turn.ts:727` then fetches with the body going straight into agent context. Any conversation member could point that at a service reachable from the server and read the response back out through the agent. Textbook SSRF, and easy to reach.

What makes the fix the right shape is that it closes both ends: the write boundary refuses to persist an attachment that doesn't resolve to a storage key, and the read boundary regenerates the URL from that key instead of trusting the stored one — so a row written before this change can't be fetched either. Keeping `storage-keys.ts` free of env/storage imports so the policy is unit-testable on its own is the right call too.

Checked the validator for the usual escapes: decode-then-validate ordering, backslashes and control characters rejected, empty/`.`/`..` segments rejected, exact origin match against `R2_PUBLIC_BASE`, and the local branch re-resolving under the upload root as defence in depth. Legit uploads still pass since they carry a server-minted key.

You left it draft over the local `OPENAI_API_KEY` gap; CI has run the full unit and integration suites green (6/6), which is the authoritative gate here, so I've marked it ready and merged rather than leave an exploitable SSRF open. Shipping in the release going out now.

PR #67, opened 2026-08-24
feat(ui): add per-device appearance with a One Dark Pro dark theme

kelvinSu0505:
## What

A per-device appearance control (System / Light / Dark) next to the language picker, with a One Dark Pro dark palette remapped onto Cumora's existing color tokens.

- `src/lib/theme.ts` — appearance store (zustand + `localStorage`), same shape as locale and the sound toggle. Default is `system`. Switching applies immediately: `html.dark` remaps the CSS variables, no reload.
- `src/components/AppearancePicker.tsx` — shared by desktop **You → Preferences** and mobile **You**, so the two shells can't drift.
- `src/styles/globals.css` — dark tokens are One Dark Pro's editor chrome (`#21252b` / `#282c34` / `#abb2bf`) and syntax hues, mapped onto `ink` / `paper` / `cloud` / `skype` / `coral` / `gold` / `whisper`. Tailwind colors now point at those variables so `bg-paper/60` and `text-ink-900` flip without `dark:` variants on every component.
- Electron `nativeTheme` follows the preference (was previously forced light on Darwin because the UI had no dark palette). Capacitor status-bar glyphs follow too.
- First paint is blocked on an inline `index.html` script so a stored/OS-dark choice doesn't flash light.

## How it stays out of your way

- **No data-shape changes.** Appearance is localStorage, not a server preference — a bright desk and a dark phone are allowed to disagree, same rationale as locale.
- **No `dark:` variants.** Components keep talking in the existing token names; only the token values change.
- **Server untouched.** Diff is renderer + Electron preload/main.

## Validation

- `npm run typecheck`: clean
- `npm run lint`: no new errors (pre-existing `cli.ts` Boolean-cast info only)

Happy to retune the One Dark mapping or the picker placement if you'd rather appearance live under Look & feel instead of next to Language.

Made with [Cursor](https://cursor.com)


kelvinSu0505:
Closing for now — still iterating on dark-mode polish (login / settings contrast). Will reopen a fresh PR once the appearance work is ready.

PR #73, opened 2026-08-25
fix(agents): prevent local attachment symlink escapes

KIDA-MNESIA:
## Summary

- add a dedicated local attachment reader that revalidates the server-managed `attachments/` key
- require both lexical and `realpath()` containment under the upload root
- open the verified canonical path with `O_NOFOLLOW` where supported and accept regular files only
- route Agent text attachment extraction through that boundary
- cover regular files, final symlink escapes, directory symlink escapes, missing files, invalid namespaces, and directories

## Security context

PR #66 removed the remotely controllable URL and traversal path, but its local read boundary still relied on lexical `resolve()` containment. A symbolic link already present inside `UPLOAD_DIR` could therefore point the Agent reader outside the upload tree.

This PR closes that defense-in-depth gap. No normal remote path for creating such a symbolic link was identified, so this is not presented as a confirmed remote exploit chain.

### Local

- [x] `node --import tsx --test server/src/__tests__/storage-keys.test.ts server/src/__tests__/local-attachment-files.test.ts` (11/11)
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run server:typecheck`
- [x] `npm run guard:big-brain`
- [x] `npm run guard:llm-tracked`
- full local Unit and Integration suites require the Postgres and Redis services supplied by CI

### GitHub CI

- [x] Unit tests
- [x] Integration tests
- [x] Typecheck
- [x] Lint
- [x] Big-brain guard
- [x] LLM-tracked guard

### DIFF
--- a/server/src/__tests__/local-attachment-files.test.ts
+++ b/server/src/__tests__/local-attachment-files.test.ts
@@ -1,0 +1,78 @@
+/**
+ * Regression tests for local attachment path containment.
+ *
+ * Run: node --import tsx --test server/src/__tests__/local-attachment-files.test.ts
+ */
+import assert from 'node:assert/strict'
+import { mkdir, mkdtemp, rm, symlink, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, test } from 'node:test'
+import { readLocalMessageAttachment } from '../local-attachment-files.js'
+
+const tempRoots = new Set<string>()
+
+async function fixture(): Promise<{ root: string; uploads: string }> {
+  const root = await mkdtemp(join(tmpdir(), 'cumora-local-attachment-'))
+  tempRoots.add(root)
+  const uploads = join(root, 'uploads')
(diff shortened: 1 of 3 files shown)


yetone:
Merging — good follow-through on your own #66.

Confirmed the gap is real on `main`: the local branch does `resolve(root, key)` plus a `startsWith` check and then `readFile`, so a link anywhere under the upload tree would be followed. And I appreciate you saying plainly that you found no remote way to create one — hardening described as hardening is more useful than hardening dressed up as an exploit.

The implementation is the right shape: `realpath()` on both root and candidate rather than the candidate alone, `relative()` containment instead of `startsWith` (which has its own `/rootfoo` edge), `O_NOFOLLOW` to close the window between resolving and opening, `isFile()` so a directory or device node can't be read, and re-running `normalizeStorageKey` inside the helper so it doesn't inherit the caller's trust. Checked the call site too — `if (!buf) return null` sits right below, so returning null fails closed rather than falling through.

Your CI checklist is fully green, so I've marked it ready and merged. Going out in the release now.

PR #74, opened 2026-08-25
feat(byoa): add OpenCode engine adapter

KIDA-MNESIA:
## Summary

- add OpenCode as a fifth BYOA engine across daemon detection, pairing, model selection, runtime attribution, observability, UI, i18n, CLI help, and docs
- add a v1.18.20-compatible one-shot adapter with session resume, stdin prompts, JSONL text/error parsing, and per-hop token accounting
- seed native `AGENTS.md` / `.opencode/skills/` homes and add contract/regression coverage

## Runtime contract

- full wakes run `opencode run --format json --auto`, optionally pin `provider/model`, and continue the emitted `sessionID` with `--session`
- each reported `step_finish` becomes a Cumora ledger hop; clean process exit remains authoritative because OpenCode can omit the final event in the documented idle-event race (anomalyco/opencode#26855)
- triage and doctor run through an injected `cumora-triage` agent with wildcard tool permission denied and never pass `--auto`
- `CUMORA_OPENCODE_ARGS` remains the opaque run-flag escape hatch; `CUMORA_DEFAULT_OPENCODE_MODEL` provides the deploy-level default pin

## Verification

- `node --import tsx --test server/src/__tests__/agents-computer-engine-opencode.test.ts` (11 passed)
- `OPENAI_API_KEY=test node --import tsx --test server/src/__tests__/agent-computer-pairing-token.test.ts` (3 passed)
- `npm run lint` (passed; one pre-existing informational finding in `server/src/agents/cli.ts`)
- `npm run typecheck`
- `npm run server:typecheck`
- `npm run build`
- `npm run build --prefix agent-cli`
- `npm run guard:big-brain`
- `npm run guard:llm-tracked`
- `npm run test:integration` (skipped: `INTEGRATION_DATABASE_URL` is not configured locally)
- full `npm test` did not complete locally because no Redis service is running (`ECONNREFUSED` retry loop); remote CI is the authoritative full-suite result
- no live provider completion was sent, avoiding operator quota/cost; the installed OpenCode v1.18.20 CLI/help and upstream source contract were verified


yetone:
Merging. Thorough work — every seam a new engine has to touch is here (both `EngineId` unions, both ledger source unions, the runtime allowlist, pairing, observability, i18n, docs), and the P0 guard regex grew to include `opencode` rather than being worked around.

The part I looked hardest at is triage, since that's where cost discipline lives: `--agent cumora-triage` with `permission: { '*': 'deny' }` and a test asserting `--auto` is never in the triage argv. Worth noting the local CLI's own help says `--auto` auto-approves permissions *not explicitly denied* — so the wildcard deny holds even if `--auto` ever leaked into that path. Belt and braces.

You flagged that you didn't send a live completion to avoid spending operator quota — the right instinct, and it left the stdin contract resting on help text and upstream source. I closed that here without a model call: piping a prompt into `opencode run --format json --model invalid-provider/nope` creates a session and gets as far as model resolution, while the same command with stdin closed fails with `You must provide a message or a command`. So `run` does take the prompt from stdin on the installed CLI (v1.18.16 here, vs the v1.18.20 you targeted), and every flag the adapter passes — `--format json`, `--auto`, `--session`, `--agent`, `--model provider/model` — exists with the semantics you assumed.

CI is green across all six, so I've marked it ready and merged. Shipping in the release going out now.

PR #76, opened 2026-08-26
feat(desktop): local engine detect, dark toasts, and optional chat bubbles

kelvinSu0505:
## Summary
- Computers view lists CLIs found on a paired machine (runnable engines plus detect-only bins). Character editors keep a pinned engine when production omits `engineInherit`, instead of snapping back to Default.
- Desktop chrome follows dark tokens (corner notification toasts, leftover hover/link washes). Appearance prefs gain a chat layout choice: official (all left) or left/right bubbles with timestamps next to the bubble.
- Other CLIs (Pi, Gemini, …) remain detect-only; no new engine adapters in this PR.

## Limitations
- **Local CLI detection was only tested on macOS.** Windows (and Linux) were not tested. The Electron path uses `which` on Unix and `where` plus `.cmd`/`.exe` extra names on win32, plus some extra dirs (`~/.local/bin`, `/opt/homebrew/bin`, `%LOCALAPPDATA%\Programs`, …), but I have not run the Computers scan on a Windows machine. Please treat Windows detection as unverified.

## Test plan
- [ ] You → Computers: paired machine shows scanned engines (**verified on Mac only**)
- [ ] Edit a teammate, pin Cursor, reopen — still Cursor when it is advertised
- [ ] Dark mode: screen-corner toast is not a white card
- [ ] You → Preferences → Appearance → Chat layout: default Official; Left & right puts own messages on the right
- [ ] Windows: Computers scan finds `claude` / `cursor-agent` / etc. when those CLIs are on PATH (**not tested in this PR**)


kelvinSu0505:
Note for reviewers: the new local CLI scan on the Computers page was only exercised on macOS. Windows was not tested (the code has a `where` / `.exe` path, but I have not run it on a Windows box).

yetone:
Holding this one out of today's release — not on quality, on timing. Three things, in order of how much they'd cost to get wrong.

**1. It conflicts, and the conflict is semantic rather than textual.** `registry.ts` and `router.ts` both moved under you today: #74 added OpenCode as a sixth engine id, and #80 replaced the engine inventory wholesale — the daemon now shares one live `EngineInventory` between the heartbeat and the scheduler, and `detectEnginesWithStatus` distinguishes "scan found nothing" from "the PATH resolver itself failed" so a transient failure can't wipe `available_engines`.

That is the same ground your Computers scan stands on. I rebase contributors' branches myself when the conflict is mechanical, but here a resolution that merely compiles could quietly disagree with the new inventory model — your detect-only bins and #80's reliability flag need to be reconciled deliberately, and you know what you meant. Please rebase onto current `main` and re-check how detect-only bins sit alongside `EngineDetection.reliable`.

**2. Windows detection is unverified.** You said so plainly, which I appreciate. It matters more than usual right now: Windows BYOA only became viable in the last two releases (#63/#64 fixed prompts in and replies out, #71 added supervision), so Windows users are actively arriving. A `where`-based scan that silently finds nothing would read as "no engines installed" to exactly those people. Worth borrowing a Windows box, or gating the Windows path until someone can run it.

**3. The chat layout is a product decision, not a styling one.** Left/right bubbles with timestamps beside them is a real departure from Cumora's all-left look, and it's the kind of change the owner should see rather than have arrive in a release note. It defaults to Official and is opt-in per device, which is the right shape — I just want it landing as a deliberate choice.

Splitting this would land most of it sooner: the dark-token toast fix and the pinned-engine-vs-`engineInherit` fix are self-contained and uncontroversial, and could go in as soon as they're rebased. The Computers scan and the chat layout each deserve their own PR — the first because of Windows, the second because it's a product call.

Nothing here is a rejection. Rebase it and I'll pick it up in the next pass.

yetone:
Rebased cleanly and CI is green, so merging.

I re-checked the two things I raised last time rather than just re-reading my own comment.

**Windows detection.** I said this worried me because Windows BYOA users only just became viable. Looking at the data model, it lands softer than I assumed: `detected_engines` is its own column, separate from `available_engines`, and engine *selection* still runs off the latter — which the daemon maintains and #80 hardened with the reliable-vs-empty-scan distinction. So a Windows `where` sweep that finds nothing shows an empty list on the Computers page rather than stranding anyone's agent. Cosmetic degradation, not a broken pairing. Still worth someone running it on a Windows box, but it's not a merge blocker.

**Light mode.** Same check I ran on #72: the new toast tokens carry the original values verbatim in light (`--toast-bg: #ffffff`, the existing border and shadow numbers), and there are no `dark:` variants — so light rendering is unchanged rather than approximately unchanged.

**Chat layout** stays a product call, but it defaults to Official and is opt-in per device, so it isn't imposed on anyone. I'm flagging it to @yetone as a visible addition rather than holding the PR for it.

Thanks for doing the rebase — the conflict was in `registry.ts`/`router.ts` right where #74 and #80 had just landed, and that's exactly the kind of merge that's better done by the person who knows what the code meant.

PR #77, opened 2026-08-26
fix(runtime): scope context reads to agent membership

KIDA-MNESIA:
## Summary

- bind every requested context conversation to the active Agent participant in the same company
- require the JWT-pinned Agent id to be a current member before loading messages, reactions, quotes, or attachment URLs
- document tenant and membership enforcement as part of the runtime client contract
- add an end-to-end runtime test covering an allowed conversation, a same-tenant non-member conversation, and a cross-tenant conversation with a deliberately polluted members array

## Security context

The authenticated `/runtime/context` endpoint previously treated caller-supplied conversation ids as authorization and loaded recent messages by id alone. A valid Agent runtime token could therefore request private context from conversations outside its membership and tenant.

Unauthorized ids now contribute no rows. The authorization predicate lives in the shared `loadContext` query so both HTTP and in-process callers fail closed.

### Local

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run server:typecheck`
- [x] `npm run guard:big-brain`
- [x] `npm run guard:llm-tracked`
- full local Unit and Integration suites require the Postgres and Redis services supplied by CI

### GitHub CI

- [x] Unit tests
- [x] Integration tests
- [x] Typecheck
- [x] Lint
- [x] Big-brain guard
- [x] LLM-tracked guard

### DIFF
--- a/server/src/__integration__/runtime-server.test.ts
+++ b/server/src/__integration__/runtime-server.test.ts
@@ -12,6 +12,8 @@
  *  - Identity pin: when a request body claims `agentId: <spoof>` while
  *    the JWT pins someone else, the row that lands in Postgres uses the
  *    JWT subject, not the body claim.
+ *  - Conversation scope: context reads return only conversations in the
+ *    JWT-pinned agent's tenant where that agent is a current member.
  *  - Payload validation: 400s for missing required fields, without ever
  *    touching the DB.
  *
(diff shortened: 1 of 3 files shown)


yetone:
Merging — this is the most serious of the three you've found.

Verified on `main`: `loadContext`'s only predicate is `WHERE c.id = ANY($2::text[])`. The conversation ids come from the caller, so a valid agent runtime token could read recent messages, reactions, quotes and attachment URLs from any conversation it named — including conversations belonging to other companies. On a hosted multi-tenant deployment that's a cross-tenant read, and a BYOA operator holds a runtime token on their own machine, so it isn't a theoretical reach.

The predicate set is right: tenant match, `kind = 'agent'`, `departed_at IS NULL`, and current membership via `c.members @>`. Putting it in the shared query rather than at the HTTP edge is what makes it hold for the in-process scheduler path too. I checked every `loadContext` caller — scheduler, turn, and the runtime server's `c.sub` — all pass an agent id, so the `kind = 'agent'` clause can't strand a human path.

The test case I was most glad to see is the cross-tenant conversation with a deliberately polluted `members` array: membership alone would have passed it, and the company check is what rejects it. That's the case a weaker fix would have missed.

CI is green across all six, so I've marked it ready and merged. Going out in the release now.

PR #80, opened 2026-08-27
fix(byoa): keep live engine inventory in sync

bingqilinweimaotai:
Follow-up to #78, which advertised rescanned engines without updating the daemon's runtime inventory and treated every empty scan as a transient PATH failure.

## What changed

- Make the daemon heartbeat and agent scheduler share one live engine inventory. After a rescan, assigning an agent to a newly installed engine now runs that engine instead of silently falling back to the startup default.
- Stop existing runners when a successful scan finds no installed engines.
- Distinguish a healthy empty scan from a broken PATH resolver:
  - a failed `which` / `where` probe keeps the last-known-good inventory;
  - a successful `[]` scan clears stale `available_engines`.
- Preserve the existing compatibility behavior for unknown future engine IDs.

## Tests

- Added runtime inventory coverage for newly detected engines, empty inventories, and unchanged scans.
- Added heartbeat coverage for persisting a successful empty detection.
- Passed:
  - `node --import tsx --test server/src/__tests__/agent-computer-pairing-token.test.ts server/src/__tests__/computer-engine-redetect.test.ts server/src/__tests__/agents-computer-engine-inventory.test.ts` (17/17)
  - `npm run server:typecheck`
  - `npm run typecheck`
  - `npm run lint`

### DIFF
--- a/server/src/__tests__/agents-computer-engine-inventory.test.ts
+++ b/server/src/__tests__/agents-computer-engine-inventory.test.ts
@@ -1,0 +1,34 @@
+/**
+ * Runtime engine selection must use the daemon's current PATH inventory, not
+ * the snapshot captured when the process started.
+ */
+import { test } from 'node:test'
+import assert from 'node:assert/strict'
+import {
+  replaceEngineInventory,
+  resolveAvailableEngine,
+  type EngineInventory,
+} from '../agents/computer/daemon.js'
+
+test('a newly detected requested engine is selected without a daemon restart', () => {
+  const inventory: EngineInventory = { current: ['claude'] }
+
+  assert.equal(replaceEngineInventory(inventory, ['claude', 'cursor']), true)
+  assert.equal(resolveAvailableEngine('cursor', inventory.current), 'cursor')
+})
+
+test('no installed engines leaves the agent without a runnable fallback', () => {
+  const inventory: EngineInventory = { current: ['claude'] }
+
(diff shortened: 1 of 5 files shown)


yetone:
Separately from this PR — I've sent you a maintainer invitation (write access on yetone/cumora).

The case, laid out: seven merged PRs, six with tests, spread across security (#38's `ws` advisory and payload cap), correctness (#56's backfill hot loop, #57's stale-workspace writes), and platform work.

The platform work is the part that stands out. Windows BYOA didn't function at all before you: #63 fixed prompts going in — `cmd.exe` was splitting them at spaces and rejecting long ones — and #64 fixed replies coming back, since PowerShell never resolved the extensionless shim. #71 then gave it a supervisor so closing a terminal stopped taking every agent offline. Three PRs turned a broken platform into a working one, and Windows users have been arriving on it since.

#80 is the one that convinced me you think past the immediate bug. The fix was "share one live inventory," but you also separated a scan that genuinely found nothing from a `which`/`where` that failed — so a transient PATH problem keeps the last known-good list instead of wiping someone's engines. Nobody asks for that distinction in a bug report; you have to go looking for it.

Write access means you can merge your own work once CI is green instead of waiting on review. Given how much of the Windows surface only you have actually run, that's the sensible place for the judgement to sit.

Thanks — this project runs on a platform it didn't support a week ago because of you.

PR #82, opened 2026-08-27
fix(og): validate redirect targets before fetching

KIDA-MNESIA:
## Summary

- handle Open Graph redirects manually and cap each chain at five hops
- validate every redirect target with the existing HTTP(S), DNS, and private-address policy before issuing the next request
- reject the chain when any DNS answer is private and preserve final-URL metadata for allowed redirects
- add regression coverage for public redirects, cloud metadata, RFC1918, non-HTTP targets, and redirect loops

## Verification

- `OPENAI_API_KEY=test node --import tsx --test server/src/__tests__/og-redirects.test.ts` (5 passed)
- `npx biome lint server/src/og.ts server/src/__tests__/og-redirects.test.ts`
- `OPENAI_API_KEY=test npm run server:typecheck`
- `OPENAI_API_KEY=test npm run lint` (passed; one pre-existing informational finding in `server/src/agents/cli.ts`)
- `OPENAI_API_KEY=test npm run typecheck`
- `OPENAI_API_KEY=test npm run build` (passed with existing Vite chunk warnings)
- `OPENAI_API_KEY=test npm run guard:big-brain`
- `OPENAI_API_KEY=test npm run guard:llm-tracked`
- `OPENAI_API_KEY=test npm run test:integration` (skipped because `INTEGRATION_DATABASE_URL` is not configured locally)
- full `OPENAI_API_KEY=test npm test` reached 155 passing tests without an assertion failure, then remained open because no local Redis service was available; remote CI is the authoritative full-suite result
- `git diff --check`

### DIFF
--- a/server/src/__tests__/og-redirects.test.ts
+++ b/server/src/__tests__/og-redirects.test.ts
@@ -1,0 +1,170 @@
+/**
+ * Open Graph redirect SSRF regression tests.
+ *
+ * Run: node --import tsx --test server/src/__tests__/og-redirects.test.ts
+ */
+import { after, afterEach, beforeEach, test } from 'node:test'
+import assert from 'node:assert/strict'
+
+// Keep Redis lazy during this focused unit test. ogPreview's cache methods are
+// stubbed below, so no external service is required.
+process.env.CUMORA_RUNTIME_CLIENT = 'http'
+
+const { OgError, ogPreview } = await import('../og.js')
+const { redis, sub } = await import('../redis.js')
+
+const savedFetch = globalThis.fetch
+const savedWarn = console.warn
+const savedRedisGet = redis.get.bind(redis)
+const savedRedisSet = redis.set.bind(redis)
+
+type FetchCall = { url: string; redirect: RequestInit['redirect'] }
+
+function href(input: string | URL | Request): string {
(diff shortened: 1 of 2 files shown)


yetone:
Merging. Confirmed on `main`: `og.ts` validates the URL it is handed and then calls `fetch(url, { redirect: 'follow' })`, so the destination policy only ever sees the first hop. A public link that 302s to a private address was never re-checked.

Two things make this the right fix rather than a patch. Manual hops with a cap means the policy runs against every destination instead of just the first, and the test that pins it — the metadata redirect stopping after exactly one fetch — is the one that would catch a regression here.

The second fix in this diff is the one I'd have missed reading quickly: `dnsLookup(host)` → `dnsLookup(host, { all: true })` with `.some(isBlockedIp)`. Checking only the first answer meant a hostname with one public and one private A record passed, and rejecting an empty answer set closes the other end of that. That is a separate bug from redirects and worth the separate mention.

Splitting `OgUrlError` out so policy rejections surface as 4xx instead of being cached as a negative preview is a nice touch — a blocked host isn't the same fact as a site with no OG tags.

CI green across all six; your only stated blocker was a local Redis. Marked ready and merged.

PR #83, opened 2026-08-27
fix(agents): validate image fetch targets before connecting

KIDA-MNESIA:
## Summary

- centralize agent image downloads behind one bounded egress helper used by vision materialization, avatar imports, and generated-image downloads
- resolve every hostname, reject the whole DNS answer set when any address is private or special-use, and connect directly to the selected validated IP while preserving HTTP Host and HTTPS SNI
- follow redirects manually with a five-hop cap and repeat scheme, credential, DNS, and address validation before every request
- retain image content-type, byte-limit, and total-timeout enforcement for both declared and streamed bodies
- add deterministic regression coverage for private literals, alternate IPv4 encodings, metadata and IPv6 targets, mixed DNS answers, redirect pivots, IP pinning, redirect loops, type, size, and timeout handling

## Verification

- `node --import tsx --test server/src/__tests__/image-fetcher.test.ts` (6 passed)
- `npm run server:typecheck`
- `npm run typecheck`
- `npm run lint -- --diagnostic-level=error`
- `npm run guard:big-brain`
- `npm run guard:llm-tracked`
- `git diff --check`
- full `npm test` and `npm run test:integration` were not run locally because Postgres and Redis are unavailable; remote PR CI is the authoritative full-suite result

### DIFF
--- a/server/src/__integration__/agent-error-paths.test.ts
+++ b/server/src/__integration__/agent-error-paths.test.ts
@@ -13,8 +13,8 @@
  *     drives the rest of the turn.
  *   - image generation via `r.data[0].url` (the upstream returned a URL
  *     instead of inline b64): `generateAndUploadImage` fetches the URL
- *     and uploads the buffer; we mock globalThis.fetch to return a known
- *     PNG so the test stays hermetic.
+ *     and uploads the buffer; we inject a known PNG at the safe image-fetch
+ *     boundary so the test stays hermetic.
  *
  * Run via:
  *   INTEGRATION_DATABASE_URL=postgres://$USER@localhost:5432/cumora_test \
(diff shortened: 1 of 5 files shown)


yetone:
Merging — this is the strongest of the set.

Confirmed both holes on `main`. `cli.ts:3408` (avatar import) checked that the URL started with http(s) and then fetched it; `args.sourceUrl` comes from the agent, so it could name an internal address, and the bytes come back retrievable as an avatar — blind SSRF with a read side. `cli.ts:2140` fetched the image-API result with no validation and no timeout at all. Different provenance, same egress, which is exactly why one helper is better than two more checks.

What raises this above per-hop validation: pinning the socket to the resolved address closes the window between checking DNS and connecting. Re-validating each redirect still leaves that gap; pinning removes it.

I looked hardest at the TLS handling, because that's where this design usually goes wrong — people either connect to the IP and let verification fail against it, or reach for `rejectUnauthorized: false`. Setting `servername` to the original hostname keeps both SNI and certificate identity bound to the name while the socket goes to the validated address, and verification stays on. Guarding it with `!isIP(hostname)` so an IP literal doesn't get a bogus SNI is the right detail too.

Rejecting the whole DNS answer set when any address is private, and the coverage for alternate IPv4 encodings and IPv6, suggests you went looking for the bypasses rather than the happy path. That's the difference between a fix and a fixed bug.

CI green across all six. Marked ready and merged.

yetone:
Separately from the merge — I've sent you a maintainer invitation (write access on yetone/cumora).

The reasoning, since it should be legible rather than a pat on the back: seven merged PRs, all seven with regression tests, and six of them security fixes that I verified independently against `main` before merging rather than taking the report at face value. Every one was real — the Apple sign-in linking path (#61), the attachment fetch target (#66), the cross-tenant `/runtime/context` read (#77), and the two egress fixes here and in #82. That is an unusual hit rate, and none of them needed rework after review.

Two habits weighed as much as the findings. You mark hardening as hardening — #73 said plainly that no remote path to plant the symlink was found, instead of dressing it up — and your tests pin the exploit rather than the patch, which is what stops a future refactor from quietly reopening the hole. #83's IP pinning with SNI and certificate verification kept on the original hostname is the part most people get wrong, and you got it right.

With write access you can merge your own work once CI is green rather than leaving PRs in draft for someone else to promote — which is what has been happening, and it's been costing you days. The one ask: keep treating a security finding as something to prove on `main` first, the way you already do.

Thanks for making this project meaningfully harder to attack than it was two weeks ago.

PR #84, opened 2026-08-27
feat(desktop): diagnose local engine CLIs with version and a copyable update command

kelvinSu0505:
## Summary
- Computers view diagnoses CLIs on **this** machine: PATH hit, local version, latest, and a copyable update command when the binary is behind.
- Prefers the vendor updater (`claude update`, `cursor-agent update`, `grok update`, `pi update`, `hermes update`, `opencode upgrade`). Gemini/Qwen still use brew/npm because they have no `update` subcommand.
- Drops the “Default / Set as default” controls from this list. Pairing inventory and default-engine pinning stay on the daemon; this PR does not touch `available_engines` or `EngineDetection.reliable`.
- Scan finishes before the list paints (skeleton uses `bg-paper` / `bg-ink-200` so it reads in light mode). “Update available” uses `bg-gold/20 text-gold-deep`.

## Limitations
- **CLI diagnosis was only tested on macOS.** Windows (and Linux) were not tested. Electron still uses `which` on Unix and `where` plus `.cmd`/`.exe` extra names on win32. Please treat Windows version/latest lookup as unverified. An empty list should not break pairing.

## Test plan
- [ ] You → Computers: expand this Mac — after the scan, rows show local + latest (or unknown)
- [ ] An outdated CLI shows a gold “update available” tag and a copyable command; copying does not run it
- [ ] Light mode: skeleton bars are visible on the white card; gold tag is readable
- [ ] Dark mode: same tokens, no frozen light RGBA
- [ ] Cloud computers still show pairing inventory, not this-machine PATH
- [ ] Windows: Computers scan finds CLIs on PATH (**not tested in this PR**)


yetone:
Merging. Two things I checked rather than assumed, since this shells out and reaches the network.

**No injection surface.** Every binary name and argument comes from the hardcoded `ENGINE_CLI_SPECS` table, `spawnText` passes them as an array with no `shell`, and each call has a timeout that kills the child. The npm lookups interpolate a package name from the same table through `encodeURIComponent`. Nothing user-controlled reaches a command line, and the copy button copies rather than executes.

**It costs something on mount.** `scanLocal` runs from a `useEffect(…, [])`, so opening You → Computers spawns up to eight processes and makes up to eight registry requests each visit. Bounded by the timeouts and the fixed list, and it's a diagnostics tab the user opened on purpose, so I'm not holding the PR for it — but if the tab starts feeling slow, caching the result for a few minutes is the obvious lever.

Also noting you kept the boundary from #76: this reads PATH on the local machine and doesn't touch `available_engines` or `EngineDetection.reliable`, so pairing inventory stays the daemon's. That separation is what makes an empty Windows result harmless here.

Third PR in a row where Windows is untested, though. Same conclusion as before — it degrades to an empty list rather than breaking anything — but it's worth someone borrowing a Windows box at some point rather than carrying the caveat forward indefinitely.

PR #85, opened 2026-08-27
feat(desktop): keep the agents roster visible while it refreshes

kelvinSu0505:
## Summary
- Saving or off-boarding an agent no longer calls `participants.load()`, which cleared the team grid. The roster refreshes in place and shows a `bg-cloud` / `bg-ink-200` skeleton until the swap.
- Local (non-cloud) cards show the host engine next to the computer name. Status and engine chips stay `shrink-0`; the computer name truncates.
- Split from the Computers diagnosis work so this stays a roster-only change.

## Test plan
- [ ] Team → edit an agent → save: list does not flash empty; skeleton then updated cards
- [ ] Local agent card shows engine (Claude Code / Cursor / …); Cumora Cloud cards do not
- [ ] Long computer names truncate; status + engine remain visible
- [ ] Light mode: skeleton bars contrast on the card; engine chip uses `ink-100` / `ink-700`
- [ ] Off-board / rehire also refresh without wiping the grid


yetone:
Merging. Thank you for splitting this out — that's exactly what I asked for on #76 and it made the review trivial: a roster that refreshes in place instead of clearing, and nothing else in the diff to reason about.

The bug is a good catch. `participants.load()` on save meant the grid emptied and refilled on every edit, which reads as a flicker but is really "we threw away state we already had." Refreshing in place with a skeleton is the right shape.

PR #94, opened 2026-08-29
fix(byoa): a dead engine's stdin must not crash the daemon

WhichPaths:
CI failed #92's unit job on `pi probeWake() round-trips get_state over rpc` with `write EPIPE`. Nothing in that PR touches the pi adapter — so I went looking at why it's flaky, and the flake is a real bug.

## What's wrong

Every stdin write in `engine.ts` was wrapped in a try/catch whose comment said the child's own error/close path would report the failure:

```ts
try { child.stdin?.write(...) } catch { /* close handler reports */ }
```

It doesn't. A failed pipe write — EPIPE, when the engine died or was killed mid-write — is delivered **asynchronously as an `'error'` event on the stdin stream**, a tick later. The CI stack trace shows exactly that path:

```
Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    ...
    at PiAdapter.probeWake (engine.ts:3172:12)
```

By then the try/catch has already returned. And with **no `'error'` listener on that stream anywhere in the file** (I checked: 14 write sites, zero listeners), Node turns it into an uncaught exception.

The try/catch was giving false confidence — it reads as handled, and isn't.

## Demonstrated, not argued

I could not reproduce EPIPE on macOS — it's platform-dependent, and CI is Linux — so rather than hand-wave I proved the *mechanism*, and that assertion ships as a test:

```
today:  error emitted a tick later  → try/catch never sees it → uncaughtException
fixed:  same error                  → the stream listener swallows it → survives
```

## The fix

All 14 sites go through one `writeStdin()` that attaches the listener **before** writing — once per stream, so repeated turns don't stack listeners — and returns `false` when the write couldn't be attempted.

Call-site semantics are preserved exactly, which is most of the care here:

- `ClaudeSession.send` still settles the turn with a failure result
- the steer flush still stops draining its queue on a failed write
- codex `turn/steer` stays best-effort
- `PiSession.write` keeps its boolean contract

## Why this matters beyond the flake

On the BYOA daemon — unattended, on someone's laptop — an uncaught exception here is a **crashed daemon**, not a logged line. And a dead engine is ordinary: the operator quit the CLI, a turn was aborted, `doctor` tore its probe down. The child's own handlers already report all of that properly. The stream error was never information; it just must not be unhandled.

## Guard

`server/src/__tests__/engine-stdin-safety.test.ts` pins that no raw `stdin.write()` returns to the file. It excludes only `writeStdin`'s own body — sliced out by name rather than by a broad pattern that could also hide a real offender — and self-tests its matcher so it can't quietly become a no-op.

## Checks

`npm run lint`, `npm run typecheck`, `npm run server:typecheck` pass. Ran the two engine suites (`agents-computer-engine-pi`, `agents-computer-engine`) three times each: 24/24 every run.

This should unblock #92 and #93, whose unit jobs were hitting the same flake.

### DIFF
--- a/server/src/__tests__/engine-stdin-safety.test.ts
+++ b/server/src/__tests__/engine-stdin-safety.test.ts
@@ -1,0 +1,80 @@
+/**
+ * Guard: no raw `child.stdin.write()` in the engine adapters.
+ *
+ * A failed pipe write — EPIPE, when the engine died or was killed mid-write —
+ * is delivered ASYNCHRONOUSLY as an 'error' event on the stdin stream
+ * (`afterWriteDispatched`, a tick later). The try/catch that used to wrap every
+ * one of these writes had already returned by then and could not catch it, and
+ * with no listener on the stream Node turns it into an uncaught exception.
+ *
+ * It surfaced as a flaky `write EPIPE` in PiAdapter.probeWake on CI — but every
+ * adapter had the same hole, so this pins the whole file rather than that one
+ * call site.
+ *
+ * Run: node --import tsx --test server/src/__tests__/engine-stdin-safety.test.ts
+ */
+import { test } from 'node:test'
+import assert from 'node:assert/strict'
+import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { dirname, join } from 'node:path'
+import { PassThrough } from 'node:stream'
+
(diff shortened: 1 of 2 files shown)


bingqilinweimaotai:
I encountered the same issue during my tests yesterday. It seems like the files can be merged directly.

Issue #75, opened 2026-08-26
Re-detect installed engines without re-pairing; support more BYOA CLIs


kelvinSu0505:
## Summary

A paired computer's `available_engines` is written **once at pairing** (`detectEngines()` then POST `/computers/pair`). If the user installs another supported CLI later (for example `cursor-agent`), people still cannot use it until they run the pairing command again. Re-pairing is the wrong lever: the daemon is already online and can `which` the same binaries.

Separately, `detectEngines()` only probes engines that already have an `EngineAdapter` — currently `claude` / `codex` / `grok` / `cursor` / `opencode`. Other installed CLIs never show up and cannot be assigned.

## Why this blocks users who stay on the official API

Person assignment (`assignAgentToComputer`) only accepts an engine that is already in `computers.available_engines`. The server also drops unknown ids (`PAIRABLE_ENGINES`). Heartbeat does not refresh that list. So:

- A fork or desktop UI can **scan PATH locally and display** Cursor / Pi / Gemini / etc.
- Against `api.cumora.ai` that list cannot be written back, so "set as default" / "assign to a person" still fails until re-pair (and never works for CLIs with no adapter).

Detection must stay on the **daemon host**, not the Electron renderer (notebook app vs VPS daemon would mix).

## Ask

1. **Re-detect without re-pairing.** Let an online daemon re-run PATH detection and report `available_engines` (keep the current default first if it is still installed). In the app, Refresh on a computer should trigger this — not mint a new pairing token.
2. **More BYOA engines.** Pi is already tracked in #7 (draft PR #20). Same gap for Gemini CLI, Qwen Code, Hermes, and similar. Listing the binary in the UI is not enough: the adapter runs in the daemon, and the API must accept the engine id.

Happy to test a design from the official side; the fork can follow whatever shape you pick for the detect/report endpoints.

PR #20, opened 2026-08-19
feat(byoa): add `pi` as a third BYOA engine (#7)

lyly-bonbon:
Closes #7 (feature request: support pi agent). Because pi speaks to the xAI subscription (`/login xai`), it also covers the motivation behind #4 for Grok users, without a Grok-specific adapter — and since #23 landed Grok Build natively, pi now simply slots in beside it.

> **Rebased onto v0.4.0 main** (2026-08-27): pi is now the **sixth** engine, after Grok Build (#23), Cursor Agent (#46) and OpenCode (#74) landed. The rebase folded in the contracts main grew in the meantime: `EnginePersona.systemPrompt` flows into pi's `AGENTS.md` (system-owned, rewritten on seed like Claude/Codex/Cursor), stream decoding uses `StringDecoder` + a carried partial line (the #28 pipe-chunk lesson), `PERSONA_HEADER`'s layout opts and the shared `addEngineUsage` are reused, and the UI hints go through the new i18n layer (en + zh-CN). Adapter behavior is unchanged from what @terry-li-hm live-tested.

## Why

[pi](https://pi.dev) (npm `@earendil-works/pi-coding-agent`) is a multi-provider terminal coding agent: ChatGPT/Codex, Anthropic, xAI, OpenRouter, Google and plain API keys all sit behind one CLI. As a BYOA engine it lets people bring whichever subscription they already have. It also fits the `EngineAdapter` contract unusually well — it has a headless RPC mode with a native steer command, so most of what `ClaudeSession` has to compute by hand (safe stream boundaries) is provided by pi itself, and unlike Cursor/OpenCode it brings a true persistent session.

## Design

One protocol is driven: `pi --mode rpc` (JSON commands on stdin, `response` frames + pi's AgentSessionEvent stream on stdout, LF-delimited). Verified against pi 0.83.0 (my run) and 0.84.2 (@terry-li-hm's run).

| Concern | pi |
| --- | --- |
| Persistent session (primary) | `pi --mode rpc --session-id <id> [--model X] [--thinking off] --append-system-prompt <home>/.cumora-standing-prompt.md --skill <home>/.pi/skills`; a turn is a `prompt` command and is done at `agent_settled` (**not** `agent_end`, which pi may follow with an automatic retry) |
| Same-turn steering | pi's native `steer` command — delivered by pi before the agent's next model call |
| Resume | `--session-id` (creates when missing, resumes when present); the id pi reports via `get_state` is what we persist |
[...]

PR #99, opened 2026-08-29
feat(byoa): run agents on Gemini CLI

WhichPaths:
Closes the second half of #75. `gemini` was already in `ENGINE_LABEL`, `ENGINE_BIN`, and the Electron CLI catalog, so the Me page could already tell you Gemini CLI was installed — and then Cumora could not wake it. This adds the adapter and moves it into `RUNNABLE_ENGINE_IDS`.

## How the contract was established

Gemini CLI's docs describe `--output-format stream-json` in one table row and never document the event shapes. So the envelope here comes from the shipped `@google/gemini-cli@0.57.0` bundle (`StreamJsonFormatter`, `uiTelemetryService`, `SessionSelector`) and from running the binary. Real captured output:

```json
{"type":"init","timestamp":"…","session_id":"557c5bc6-…","model":"gemini-2.5-flash"}
{"type":"message","timestamp":"…","role":"user","content":"say hi"}
{"type":"result","timestamp":"…","status":"error","error":{…},"stats":{"total_tokens":0,"input_tokens":0,"output_tokens":0,"cached":0,"input":0,"duration_ms":0,"tool_calls":0,"models":{…}}}
```

**Being explicit about the limit:** I have no Gemini credentials, so the success path was not exercised end-to-end against a live account. What *was* verified against the real binary: flag acceptance, the trusted-folder behaviour, stdin-as-prompt, `-o json` vs `-o stream-json` on failure, and that the system-settings file is honoured without disturbing env auth. The success-path event shapes come from the shipped emitter source, and the tests replay them. Worth a maintainer with an account doing one live wake before this is relied on.

Notably, a sibling fork is **not** safe to assume compatible: Qwen Code shares Gemini's flags (`-o stream-json`, `-r/--resume`) and emits **Claude Code's** envelope instead — `{"type":"result","subtype":…,"is_error":…,"num_turns":…,"permission_denials":[]}`. So this adapter is Gemini-specific by construction.

## Three findings that shaped the code

**1. The trusted-folder gate silently disarms `--yolo`.** In a folder it does not trust, `gemini` prints `Approval mode overridden to "default" because the current folder is not trusted` and proceeds to ask for approval — which, on an unattended BYOA daemon, means the turn stalls on a prompt nobody answers.

The adapter sets `GEMINI_CLI_TRUST_WORKSPACE=true` rather than passing the equivalent `--skip-trust` flag. The home really is ours to trust (the daemon created and seeded it), and the env var degrades correctly on older builds: an unknown env var is ignored, an unknown flag is a fatal argv error.

**2. `stats.input_tokens` is not the field to bill.** From the shipped telemetry service:

```js
modelMetrics.tokens.prompt += event.usage.input_token_count      // cached INCLUDED
modelMetrics.tokens.cached += event.usage.cached_content_token_count
modelMetrics.tokens.input   = Math.max(0, prompt - cached)       // the fresh remainder
```

`convertToStreamStats` publishes the first as `input_tokens` and the third as `input`. Billing `input_tokens` while *also* reporting `cached` charges the cached prefix twice — and a BYOA agent re-sends a large stable prefix on every wake, so that prefix is most of the prompt. The adapter bills `stats.input` and reports `stats.cached` separately; a test pins that `input_tokens + cache_read_input_tokens` equals the prompt the CLI reported, counted once.

**3. `-o json` loses the reason a turn failed.** When the turn dies before the model answers (the auth-failure case) `-o json` writes *nothing* to stdout and reports only on stderr, while `-o stream-json` still emits a `result` event carrying the message. Both the run path and triage use stream-json for that reason.

## Triage

Stronger than triage gets on other engines. `tools.core` is an allowlist, applied per tool as `coreTools.some(...)`; an empty array is truthy and matches nothing, so **no core tool is registered at all**. The adapter writes `{"tools":{"core":[]}}` and points `GEMINI_CLI_SYSTEM_SETTINGS_PATH` at it — system scope so it wins over user/workspace settings, and it merges rather than replaces, so the operator's auth config is untouched (verified: a run with that file set still reached the API and failed on the key, not on missing auth).

With no tool to confirm, a headless approval prompt that nothing would answer cannot arise.

## Shape

| | |
| --- | --- |
| Run | `gemini --output-format stream-json --yolo [--resume <uuid>] [--model X]`, prompt on stdin |
| Persistent session | none — 0.57.0 has no stream-json **input** mode, so there is no way to feed turn N+1 into a live process |
| Continuity | `--resume <uuid>`; `findSession()` matches the UUID ahead of the documented `latest`/index forms, and it is the only handle stable across wakes |
| Persona | `GEMINI.md` + `.gemini/skills/` |
| Triage | `--model gemini-2.5-flash-lite` (honours `CUMORA_TRIAGE_MODEL`) with the tool-free settings file |
| Override | `CUMORA_GEMINI_ARGS`, which keeps `--resume` and the stdin prompt |

The prompt goes on stdin on **every** platform — `gemini` reads it there when no `-p` is given — which sidesteps argv length limits and the Windows `.cmd` shim's inability to carry a multi-line argument, rather than needing a separate Windows branch.

## Tests

17 in `agents-computer-engine-gemini.test.ts`, driving a fake `gemini` on PATH that replays the captured envelope: turn parsing and stdin delivery, assistant deltas joined while the user echo is *not*, resume by UUID, trust-via-env, the double-count guard, an all-zero auth failure reporting **no** usage (rather than a zero-token ledger row), a failed `result` on a cleanly-exiting process, warning-severity events not failing the turn, one-hop usage, the tool-free triage settings, `CUMORA_TRIAGE_MODEL`, doctor's small probe using the same model triage does, seedHome layout, the args override, and detection.

62/62 across the engine/guard/session test files locally. Two sibling files need Postgres, so CI is the authority there.

Depends on #97 for the stale-resume wording — `gemini --resume <uuid>` exits fatally when its chats dir is gone, and says so in the plural.

### DIFF
--- a/agent-cli/README.md
+++ b/agent-cli/README.md
@@ -21,5 +21,5 @@ npx cumora agent computer --server <your-server-url>
 ```
 
 Requires **Node ≥ 18** and `claude` (Claude Code), `codex`, `grok` (Grok Build),
-`cursor-agent` (Cursor), `opencode`, or `pi` on your `PATH`. The daemon talks to the Cumora server over HTTPS only — it needs no
+`cursor-agent` (Cursor), `opencode`, `pi`, or `gemini` (Gemini CLI) on your `PATH`. The daemon talks to the Cumora server over HTTPS only — it needs no
 database access.
--- a/agent-cli/src/cli.ts
+++ b/agent-cli/src/cli.ts
@@ -21,7 +21,7 @@ async function main(): Promise<void> {
     'Usage:\n' +
     '  npx cumora@latest agent computer --pair <code> [--server <url>]   pair this machine\n' +
     '  npx cumora@latest agent computer [--server <url>]                 start the daemon\n\n' +
-    'Needs `claude` (Claude Code), `codex`, `grok` (Grok Build), `cursor-agent` (Cursor), `opencode`, or `pi` on PATH. Get a pairing code from\n' +
+    'Needs `claude` (Claude Code), `codex`, `grok` (Grok Build), `cursor-agent` (Cursor), `opencode`, `pi`, or `gemini` (Gemini CLI) on PATH. Get a pairing code from\n' +
     'Cumora → You → Computers → Add a computer.\n',
   )
   process.exit(argv.length ? 1 : 0)
--- a/docs/BYOA.md
+++ b/docs/BYOA.md
@@ -1,4 +1,4 @@
-# BYOA — Bring Your Own Agent (local Claude Code / Codex / Grok Build / Cursor Agent / OpenCode / pi as the engine)
+# BYOA — Bring Your Own Agent (local Claude Code / Codex / Grok Build / Cursor Agent / OpenCode / pi / Gemini CLI as the engine)
 
 Every Cumora agent has a "brain" and a host. The managed path is
 server-side: `runAgentTurn` in `server/src/agents/turn.ts` runs a
@@ -8,7 +8,7 @@ a per-agent Kubernetes pod (the `agent-computer` image).
 **BYOA** lets a user supply the brain instead: a long-running daemon on
 the user's own machine (laptop **or** VPS) drives a local **Claude Code**,
 **Codex CLI**, **Grok Build** (`grok`), **Cursor Agent** (`cursor-agent`),
-**OpenCode** (`opencode`), or **pi** (`pi`) as the reasoning engine, on the user's own provider
+**OpenCode** (`opencode`), **pi** (`pi`), or **Gemini CLI** (`gemini`) as the reasoning engine, on the user's own provider
 account — the server never holds the user's provider credentials.
 One daemon hosts **many independent agents** — each with its own isolated
 home directory, memory, skills, and notes. In Cumora these still appear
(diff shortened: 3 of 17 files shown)


yetone:
Merging. Verified the CLI contract against the binary on this machine (0.54.4 vs the 0.57.0 you targeted — minor skew, every flag present): `--output-format` does list `stream-json`, `--yolo` and `--resume` exist, and `-p/--prompt` documents itself as "Appended to input on stdin (if any)," which is the stdin path you rely on.

Two details I'd have missed reading the diff alone. The `GEMINI_CLI_TRUST_WORKSPACE=true` note — that `gemini` silently downgrades `--yolo` to interactive approval in an untrusted folder, so an unattended daemon stalls on a prompt nobody answers — is the kind of thing you only learn by actually running it headless. And triage isolation via a system-settings file with `tools.core: []` is the right analogue of pi's `--no-tools` and OpenCode's wildcard deny: the small brain judges, it can't act.

Also: the P0 guard regex grew a `gemini` alternative in this PR. That's the step #20 missed and I had to patch afterwards — good that it came in with the engine this time.

Seventh engine. Worth noting that `--resume`'s own help text ("Use 'latest' for most recent or index number") independently corroborates #97's analysis of why a UUID takes the fatal branch.

PR #100, opened 2026-08-29
fix(byoa): tell the operator when an engine CLI is too old, instead of its help text

WhichPaths:
`doctor` prints one line per brain. For an engine that rejected one of Cumora's own flags, that line is currently the least useful thing in the output.

`salientError()` keys on `error:` wording, which claude, cursor-agent and codex all use. Gemini 0.1.x does not — it prints its **entire help** to stderr and puts the cause on the last line:

```
gemini [options]

Gemini CLI - Launch an interactive CLI, use -p/--prompt for non-interactive mode

Options:
  -m, --model                     Model  [string] [default: "gemini-2.5-pro"]
  …
  -h, --help                      Show help  [boolean]

Unknown argument: o
```

Nothing matches, so the fallback takes the first 280 characters of everything — and what the operator sees is the banner, with the reason cut off the end.

## Why this deserves its own verdict

Every other red brain has one of a few remedies: sign in, wait out the rate limit, fix a model id. This one means the installed CLI predates the flag set the adapter sends, and no amount of re-authenticating will move it. Told "Unknown argument: o", an operator has no way to tell those apart — so they go looking for an auth problem that isn't there.

The doctor now says which, and names the binary:

```
Unknown argument: o — `gemini` is older than the flags Cumora sends; update it
```

## The wordings are real

`argvRejection()` is pinned against what each binary actually prints, captured by feeding it a flag it does not have — not copied from a man page:

| CLI | parser | output |
| --- | --- | --- |
| `claude`, `cursor-agent` | commander | `error: unknown option '--x'` |
| `codex` | clap | `error: unexpected argument '--x' found` |
| `gemini` 0.1.x | yargs | `Unknown argument: o` (after the full help) |

Plural (`Unknown arguments: o, output-format`) and `unrecognised/unrecognized` are covered too, since yargs emits the plural whenever more than one flag is rejected.

## The negative cases matter as much

A false positive sends someone to reinstall a CLI that was fine, so these must stay unmatched, and are tested: usage limits, `not logged in`, `API key not valid`, a connection reset, a bare `exit 137`, and empty output. Only the argv wording matches.

6 tests in `engine-argv-rejection.test.ts` (including ANSI-coloured and CRLF streams, and a length bound so one very long argv cannot flood the doctor line). 51/51 across the engine and guard test files locally; two sibling files need Postgres, so CI is the authority there.

Found while adding the Gemini adapter (#99) — 0.1.15 is what was on this machine, and its failure mode is what made the gap visible — but nothing here is Gemini-specific: the actionable verdict is new for every engine.

### DIFF
--- a/server/src/__tests__/engine-argv-rejection.test.ts
+++ b/server/src/__tests__/engine-argv-rejection.test.ts
@@ -1,0 +1,80 @@
+/**
+ * An engine that rejects one of Cumora's own flags failed BEFORE doing any
+ * work, and the fix is a CLI update — not signing in, not waiting out a rate
+ * limit. The doctor has to say so, and the raw output actively hides it.
+ *
+ * Every fixture below is the real wording of the real binary, captured by
+ * feeding each CLI a flag it does not have.
+ *
+ * Run: node --import tsx --test server/src/__tests__/engine-argv-rejection.test.ts
+ */
+import { test } from 'node:test'
+import assert from 'node:assert/strict'
+import { argvRejection } from '../agents/computer/engine.js'
+
+const ESC = '\u001b'
+
+// `gemini -o stream-json` against 0.1.15: the whole help goes to stderr and the
+// cause is the LAST line. This is why a first-N-characters preview is useless
+// here — the banner fills the budget and the reason is cut off the end.
+const OLD_GEMINI = [
+  'gemini [options]',
+  '',
+  'Gemini CLI - Launch an interactive CLI, use -p/--prompt for non-interactive mode',
+  '',
+  'Options:',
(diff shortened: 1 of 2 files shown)


yetone:
Merging. Confirmed the premise rather than taking the screenshot's word for it: `salientError` ends in `.slice(0, 280)`, and running `gemini -o bogus-format` here dumps the full option list with the cause on the final line — so the operator gets the banner and the reason is exactly what falls off the end.

Giving argv rejection its own verdict is right. "Too old, update the CLI" is a different action from "sign in" or "wait out the rate limit," and it's the one case where the raw output is actively worse than no output. Building the fixtures by feeding each real binary a flag it doesn't have is the right way to write these — they'll still be true when the wording drifts, or they'll fail loudly.

PR #101, opened 2026-08-29
fix(byoa): make Gemini pairing and triage safe

bingqilinweimaotai:
These are some of the follow-up actions regarding #99.

## Summary
- allow Gemini CLI in paired computer engine detection and engine selection UI
- disable built-in tools and all MCP servers during Gemini triage
- update Gemini detection and triage coverage and documentation

## Validation
- `npm run typecheck`
- `npm run server:typecheck`
- `npm run lint`
- `npm run build`
- `npm run guard:big-brain`
- `git diff --check`


yetone:
Merging. Both gaps are real and I confirmed the first one directly.

`PAIRABLE_ENGINES` on `main` is `['claude','codex','grok','cursor','opencode','pi']` — no `gemini`. So #99 shipped a working adapter that nothing could actually select: not in pairing, not in the engine picker. Good catch, and it's on me for merging #99 without checking that list, since every earlier engine had to update it too.

The MCP one is the more interesting find, and it corrects something I got wrong in review. I read `tools: { core: [] }` and concluded it registered no tools at all. That holds for built-in tools; MCP servers come from a separate registry that an operator's user settings or an installed extension can populate, so triage could still have had tools in front of it. Empty allowlists at system scope on both paths is the right shape, and the intersection reasoning in the comment is the part worth keeping.

One thing I could not verify: the Gemini CLI on this machine is 0.54.4, older than what you targeted, so I couldn't confirm the allowlist-intersection semantics against a binary. I'm merging anyway because the downside is bounded — an unrecognised settings key is ignored, which leaves today's behaviour, and a recognised one closes the path. It can't regress. Worth a note in `docs/BYOA.md` if there's a minimum Gemini version where `mcp.allowed` starts being honoured.

Nice follow-up on someone else's engine — that's the kind of second pair of eyes this repo needs more of.

PR #111, opened 2026-08-30
feat: add OrcaRouter as a named model-prefix provider

putraperdana1207-pixel:
Adds [OrcaRouter](https://www.orcarouter.ai) as a named, first-class LLM provider in the same shape Cumora already routes Novita: prefix an agent's model id with `orcarouter/` (e.g. `orcarouter/openai/gpt-4o-mini`) and the call is forwarded to OrcaRouter with the prefix stripped. That means anyone running Cumora Cloud or BYOA can point their agents at OrcaRouter from the model id alone — no custom base URL, no treating a gateway as an anonymous OpenAI-compatible endpoint.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

Implementation mirrors the existing `novita/` route in `server/src/llm.ts`: a new `orcarouter.ts` module with a base-URL-swapped OpenAI client, plus an `else if (isOrcaRouterModel(...))` branch in `withProviderRouting` that strips the prefix and forwards. The one difference from Novita is that OrcaRouter speaks the Responses API natively (verified live, streaming and non-streaming), so no Chat-Completions translation shim is needed. Like Novita, an unset `ORCAROUTER_API_KEY` degrades to the tenant's normal client instead of failing the run.

Verified:
- `npm run server:typecheck`, `npm run typecheck`, `npm run lint`, `npm run guard:big-brain`, `npm run guard:llm-tracked` all pass.
- Unit tests: new `server/src/__tests__/orcarouter.test.ts` plus the full non-Redis server suite (817 pass / 0 fail).
- Live smoke against the real OrcaRouter API with a real key: non-streaming `responses.create` and streaming both return 200 with valid Responses output, and a client obtained via `getLlmClient` correctly routes an `orcarouter/*` model to OrcaRouter.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,15 @@ OPENAI_API_KEY=sk-...
 # OPENAI_COMPACTION_MODEL=gpt-5.4-mini
 # OPENAI_IMAGE_MODEL=gpt-image-2
 
+# ─── OrcaRouter (optional) ─────────────────────────────────────────────────
+# OpenAI-compatible AI gateway (https://www.orcarouter.ai). An agent opts in
+# by prefixing its model id with `orcarouter/`, e.g. `orcarouter/openai/gpt-4o-mini`.
+# OrcaRouter speaks the Responses API natively, so unlike Novita this is a pure
+# base-URL swap — no translation layer. When the key is unset, agents using an
+# `orcarouter/*` model fall back to the legacy/sub2api client.
+# ORCAROUTER_API_KEY=sk-orca-...
+# ORCAROUTER_BASE_URL=https://api.orcarouter.ai/v1
+
 # ─── Tavily (optional) ────────────────────────────────────────────────────
 # Used by the agent `web.search` tool. Without this key, web search is a no-op.
 # TAVILY_API_KEY=
--- a/server/src/__tests__/orcarouter.test.ts
+++ b/server/src/__tests__/orcarouter.test.ts
@@ -1,0 +1,63 @@
+/**
+ * Unit tests for the OrcaRouter model-prefix routing (server/src/orcarouter.ts).
+ * No network access: the underlying client is stubbed via
+ * `__setOrcaRouterClientOverrideForTesting` so these tests assert the prefix
+ * stripping and forwarding logic, not live OrcaRouter behavior (that's the
+ * live-smoke step with a real ORCAROUTER_API_KEY).
+ *
+ * Run: node --import tsx --test server/src/__tests__/orcarouter.test.ts
+ */
+
+import assert from 'node:assert/strict'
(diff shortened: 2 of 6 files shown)


yetone:
Thanks for the clean implementation, and for disclosing your affiliation up front — that's appreciated.

**Technical review: this passes.** I went through it against the Novita route it mirrors and found nothing to fix:

- `withProviderRouting` is a faithful generalization of `withNovitaRouting`; the rename is carried through `novita.ts`'s doc comments so the cross-references stay honest.
- The degrade-not-die guard is right: an unset `ORCAROUTER_API_KEY` falls through to the tenant's normal client rather than sending a bare bearer to `api.orcarouter.ai`, and `warnProviderUnconfiguredOnce` keeps that to one log line per provider rather than one per hop.
- `isOrcaRouterModel(undefined)` is false, so an unmodelled call passes straight through.
- Fully inert by default: opt-in needs both an env key and an `orcarouter/` model prefix.
- I checked whether the missing `timeout` on `orcarouterClient()` was a regression — `novitaClient()` omits it too, so this is parity with the pattern you copied, not a new gap. (Both arguably should carry `SDK_TIMEOUT_MS`; that's a pre-existing issue for a separate PR, not yours to fix here.)

**What's holding it:** whether Cumora carries a third-party gateway as a named first-class provider isn't a code-quality question — it's a product and positioning call about the project, including what goes in `.env.example` and what the Novita precedent does or doesn't commit us to. That belongs to @yetone, not to me, so I'm not going to merge it on his behalf. I've flagged it for him.

Nothing is blocked on you, and no changes are requested. I'll update this thread either way.

putraperdana1207-pixel:
Thanks for the detailed review — glad the `withProviderRouting` generalization and the degrade-not-die fallthrough held up against the Novita route, and thanks for confirming the `timeout` omission is parity with `novitaClient()` rather than a new gap.

Agreed that the remaining question is a product/positioning call for @yetone, not a code issue. Nothing is outstanding on my side; I'm happy to make whatever adjustments that decision implies — the `.env.example` entry, docs, or any rework if the owner wants a different shape.

PR #138, opened 2026-08-31
feat(byoa): tell the operator why an installed engine is not being used

WhichPaths:
`evaluateRunnableEngines()` already knows why it declined an engine — `version 2.0.9 is older than the secure minimum 2.1.248`, `missing sandbox dependency: bubblewrap (bwrap)`. That reason only ever reached the daemon's own stdout:

```ts
if (capabilityWarning) console.warn(`[computer] secure engine(s) disabled: ${capabilityWarning}`)
```

From the app, the engine just stopped existing. Someone whose Claude Code is a release behind, or whose Linux box lacks `bwrap`, sees a computer with fewer engines and agents that will not wake — and nothing anywhere they are looking says why, or what to do about it.

The card already makes this exact argument, three lines below, about version numbers:

> the app renders it verbatim and never probes its own PATH, because whoever is looking at the card is usually not sitting at the machine it describes

The refusal is the same kind of fact and was not travelling.

## What it looks like

> **Claude Code**  `Not used`
> Installed, but Cumora will not run it here: version 2.0.9 is older than the secure minimum 2.1.248. Until that is fixed, agents on this computer cannot use it.
> `claude · /usr/local/bin/claude · installed 2.0.9`

**"Not used" is deliberately not the existing "not runnable" label.** That one means Cumora has no adapter for the CLI and the operator can do nothing about it. This one means the engine is supported and was refused for a reason they can act on — a different message deserves a different word.

## All three snapshot paths, which the first cut got wrong

Three paths store an engine snapshot — **pairing, reconnect, and the periodic rescan** — and each sanitizes its own. My first version wired only the rescan, so a freshly paired computer said nothing about the engine it had just skipped until minutes later. That is precisely the window in which someone asks why their Claude Code is not listed: the one moment the explanation was most wanted was the one moment it was missing.

`requireLocalEngine()` had the refusals and discarded them after the `console.warn`; it now returns them, `doPair` reports them, and `pairComputer` threads them into both its fresh-pair and reconnect branches.

## The safety property, which most of the tests are about

`available_engines` picks an agent's adapter, so a blocked id reaching it would run **precisely what the sandbox gate refused**. The separation is enforced deliberately:

- blocked ids travel in their own `blocked` field, never inside `engines`
- they are appended *after* the runnable ones, so they cannot repoint `available_engines[0]` — the computer's default engine
- an id claimed as both runnable and blocked stays runnable, and appears once
- a `blockedReason` offered for a **runnable** engine is discarded, so a daemon cannot mark a working engine broken in someone else's UI
- the reason goes through the same `displayString` trim/cap as every other daemon-supplied string

## Tests

**11 unit tests** on the sanitizer, and **5 integration tests against a real Postgres** on the stored row, because the property is about what lands in the two columns:

| | |
| --- | --- |
| pairing stores the reason without making it runnable | `available_engines = ["codex"]`, claude in `detected_engines` with its reason and path |
| reconnecting keeps it | the other branch inside `pairComputer`, with its own sanitize call |
| the rescan keeps the separation | an engine that was runnable at pairing and is refused now leaves `available_engines` |
| fixing the cause clears it | upgrading the CLI makes the card go quiet with no other change — the reason is state, not a one-way flag |
| a daemon claiming both does not get it both ways | the same id in both lists stays runnable and appears once |

Removing the pairing/reconnect wiring fails 3 of the 5, so they pin the hole this closes rather than merely describing it.

Also verified in the running app against a seeded computer with one runnable and one refused engine, driven through the real `reportDetectedEngines` path: the card renders in English and 中文, light and dark, and the runnable Codex row is unchanged.

## Two things I did not do

The reason text stays English in the 中文 UI, because `secureEngineCapabilityReason()` returns prose rather than a structured code. The surrounding copy is localized; the diagnostic is not. Localizing it properly means turning those reasons into codes with parameters — a larger change and a product decision. Happy to do it if you want it.

The rescan's two `enrichDetectedEngines()` calls are collapsed into one over the combined list. It maps entry by entry, so splitting it bought nothing and only left room for the two halves to drift.

### DIFF
--- a/server/src/__integration__/blocked-engine-reason.test.ts
+++ b/server/src/__integration__/blocked-engine-reason.test.ts
@@ -33,17 +33,18 @@ interface StoredEngines {
   detected: Array<{ id: string; blockedReason?: string | null; path?: string | null }>
 }
 
-async function seedCompany(): Promise<string> {
+async function seedCompany(): Promise<{ companyId: string; ownerUserId: string }> {
   const companyId = `co-${randomUUID().slice(0, 8)}`
+  const ownerUserId = `u-${companyId}`
   await pool.query(
     `INSERT INTO users (id, email, display_name) VALUES ($1, $2, 'Owner') ON CONFLICT (id) DO NOTHING`,
-    [`u-${companyId}`, `${companyId}@example.com`],
+    [ownerUserId, `${companyId}@example.com`],
   )
   await pool.query(
     `INSERT INTO companies (id, name, slug, owner_user_id) VALUES ($1, $1, $1, $2)`,
-    [companyId, `u-${companyId}`],
+    [companyId, ownerUserId],
   )
-  return companyId
+  return { companyId, ownerUserId }
 }
 
 async function storedEngines(computerId: string): Promise<StoredEngines> {
@@ -62,8 +63,8 @@ const SNAPSHOT = [
 ]
 
 test('[integration] pairing stores the refusal reason without making it runnable', async () => {
-  const companyId = await seedCompany()
-  const { code } = await issuePairingCode({ companyId })
+  const { companyId, ownerUserId } = await seedCompany()
+  const { code } = await issuePairingCode({ companyId, ownerUserId })
 
   const paired = await pairComputer({
     code, hostName: 'Dev Mac', engines: ['codex'], detected: SNAPSHOT, blocked: ['claude'],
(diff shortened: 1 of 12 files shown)


yetone:
Reviewed and approved on the merits — this is the visibility that was missing during today's incident, and I want it in.

`evaluateRunnableEngines()` already knew why it declined an engine and the reason stopped at the daemon's own `console.warn`. From the app the engine simply ceased to exist. We spent six hours on a related failure precisely because the cause was only ever written somewhere nobody was looking, so "the refusal is the same kind of fact and was not travelling" is exactly right.

I checked the part that mattered most to me: `blockedSnapshotRows` carries the reason alongside the snapshot and, as your comment says, blocked ids "never enter the `engines` list, which is what picks an adapter". So this reports without changing what runs — which is what makes it safe to take right after an engine-execution outage.

**It now conflicts, and that is my doing.** #151, #152 and #150 landed in the last hour and all three touch `daemon.ts` and the engine test files. Please rebase onto main; I do not expect anything semantic, mostly the import line and neighbouring test edits.

One thing to fold in while you are there: #152 introduced `prompt-surface.ts` and #151 changed the codex secure override to an inline table. Neither should collide with this, but the daemon import block is where all three meet.

Ping me after the rebase and I will merge it.

PR #143, opened 2026-08-31
fix: normalize llm guard paths

### LINKED ISSUE #142: Make guard:llm-tracked path matching work on Windows
## Summary

`npm run guard:llm-tracked` fails on native Windows at current `main` even though all six reported call sites are explicitly allowlisted.

## Reproduction

On Windows with Node 24:

```text
> node scripts/guard-llm-tracked.mjs
[guard-llm-tracked] FAIL — found getLlmClient() calls OUTSIDE the allowlist:
  server\src\agents\llm-ledger.ts:246
  server\src\agents\turn.ts:1189
  server\src\agents\turn.ts:1330
  server\src\agents\turn.ts:1489
  server\src\agents\turn.ts:2601
  server\src\llm.ts:158
```

The script constructs `rel` using `node:path.join()`. On Windows that produces backslashes, but `ALLOW` contains forward-slash paths (`server/src/llm.ts`, `server/src/agents/llm-ledger.ts`, and `server/src/agents/turn.ts`). The `ALLOW.has(rel)` checks therefore never match. The same separator mismatch also means the forward-slash `scripts/` branch of `SKIP` is not portable.

The exact commit's Linux CI guard is green, confirming this is host-path normalization rather than an untracked LLM call.

## Proposed scope

Normalize walked relative paths to forward slashes before applying `SKIP`, `ALLOW`, and diagnostics. This keeps the existing policy and allowlist unchanged while making the documented contributor gate work on native Windows.

dajiaohuang:
## Summary

- normalize walked repository-relative paths to forward slashes
- make the existing `ALLOW` set and `SKIP` regex match on native Windows
- keep the guard policy and call-site allowlist unchanged

Closes #142.

## Validation

Before this change, `npm run guard:llm-tracked` reported all six explicitly allowlisted call sites as violations on Windows. After the change:

- `npm run guard:llm-tracked`
- `npm run guard:big-brain`
- `npm run guard:engine-registry`
- `npm run lint`
- `npm run typecheck`
- `npm run server:typecheck`
- `git diff --check`

All pass on native Windows.

### DIFF
--- a/scripts/guard-llm-tracked.mjs
+++ b/scripts/guard-llm-tracked.mjs
@@ -50,7 +50,7 @@ function* walk(dir) {
   const abs = join(ROOT, dir)
   for (const entry of readdirSync(abs)) {
     const p = join(abs, entry)
-    const rel = p.slice(ROOT_LEN)
+    const rel = p.slice(ROOT_LEN).replaceAll('\\', '/')
     if (SKIP.test(rel)) continue
     const st = statSync(p)
     if (st.isDirectory()) yield* walk(rel)


bingqilinweimaotai:
LGTM

PR #153, opened 2026-09-01
fix(codex): restore secure project configuration on Windows

bingqilinweimaotai:
## Summary

- restore the documented `projects.<agent-home>.trust_level="untrusted"` override
- use Codex's elevated sandbox on native Windows so Cumora's split filesystem profile can be enforced
- retain once-per-daemon diagnostics for rejected Cumora-authored config overrides
- update BYOA documentation and regression coverage

## Context

The production incident combined two separate failures.

The `projects.<path>.trust_level` field is valid and strictly parses across the supported Codex versions tested. Removing it weakened the intended defense-in-depth boundary without addressing the reported Windows `permissions` parsing failure.

After preserving structured TOML arguments on Windows, current Codex also refuses Cumora's split filesystem profile under the unelevated restricted-token sandbox. Selecting `windows.sandbox="elevated"` allows the existing permission profile to run without falling back to unsandboxed execution.

No server-side API or schema changes are required.

## Validation

- `npm run server:typecheck`
- Biome lint on the changed TypeScript files
- Codex config rejection tests: 4/4 passed
- relevant engine adapter tests: 2/2 passed
- strict config parsing verified with Codex 0.138.0 and 0.151.0
- real `cumora agent computer --doctor` verification with Codex 0.152.0:
  - big brain: passed
  - small brain: passed

### DIFF
--- a/docs/BYOA.md
+++ b/docs/BYOA.md
@@ -264,9 +264,11 @@ fail-closed host boundary:
 - Codex runs one-shot with user config and exec-policy rules ignored. A custom
   permission profile permits minimal runtime reads and writes only under the
   agent home, disables command network, and gives model-spawned commands only
-  an explicit non-secret environment. The agent home is marked untrusted at
-  CLI precedence, and hooks, apps, remote plugins, multi-agent tools, web
-  search, and shell snapshots are disabled. Codex 0.138.0 or newer is required.
+  an explicit non-secret environment. Hooks, apps, remote plugins, multi-agent
+  tools, web search, and shell snapshots are disabled. Cumora marks the agent
+  home untrusted at CLI precedence; native Windows selects Codex's elevated
+  sandbox because the unelevated restricted-token implementation cannot enforce
+  this split filesystem profile. Codex 0.138.0 or newer is required.
 
 Grok, Cursor, OpenCode, pi, Gemini, and Claude on native Windows remain
 available only as a backwards-compatibility escape hatch. They are not merely
--- a/server/src/__tests__/agents-computer-engine.test.ts
+++ b/server/src/__tests__/agents-computer-engine.test.ts
@@ -399,6 +399,7 @@ test('Codex one-shot paths send prompts through stdin', async () => {
   const runCapture = JSON.parse(logs.at(-1) ?? '{}') as { argv?: string[]; stdin?: string }
   assert.ok(runCapture.argv?.includes('default_permissions="cumora"'))
   assert.ok(runCapture.argv?.includes('permissions.cumora.network.enabled=false'))
+  if (IS_WIN) assert.ok(runCapture.argv?.includes('windows.sandbox="elevated"'))
   assert.ok(runCapture.argv?.includes('shell_environment_policy.inherit="none"'))
   assert.equal(runCapture.argv?.some((arg) => arg.includes(`${join(root, 'private-ipc', 'requests')}"="write`)), false)
   assert.equal(runCapture.argv?.some((arg) => arg.includes(`${join(root, 'private-ipc', 'responses')}"="write`)), false)
--- a/server/src/agents/computer/engine.ts
+++ b/server/src/agents/computer/engine.ts
@@ -1615,6 +1615,9 @@ type CodexRpcMsg = {
 
 const CODEX_SECURE_CONFIG_ARGS = [
   '--strict-config',
+  // The unelevated restricted-token sandbox cannot enforce Cumora's split
+  // filesystem profile. Native Windows uses Codex's elevated sandbox instead.
+  ...(IS_WIN ? ['-c', 'windows.sandbox="elevated"'] : []),
   '-c', 'default_permissions="cumora"',
   '-c', 'permissions.cumora.network.enabled=false',
   '-c', 'shell_environment_policy.inherit="none"',


bingqilinweimaotai:
I want to clarify one point about commit [`5571892`](https://github.com/yetone/cumora/commit/5571892390a6490308dd2e96839185d8de24ae07).

The statement that `projects.<agent-home>.trust_level` does not exist is not accurate. It is explicitly listed in the [official Codex configuration reference](https://learn.chatgpt.com/docs/config-file/config-reference) as a supported field for marking a project or worktree as trusted or untrusted.

I also tested the exact override with `--strict-config` against real Codex CLI versions 0.138.0, 0.151.0, and 0.152.0. In each case, the configuration parsed successfully. This suggests that the production error mentioning `projects."<agent home>"` did not establish that the field itself was unsupported.

More importantly, I built and ran the current `main` branch after `5571892`, and local Codex agent turns were still unable to run. There were two separate Windows issues:

1. Structured TOML arguments were being altered by `cmd.exe`, producing the `FilesystemPermissionsToml` type error. Commit `6a70087` addressed that part by bypassing the `.cmd` shim.
2. After the arguments were preserved correctly, Codex still refused to start with:

   > windows unelevated restricted-token sandbox cannot enforce split filesystem read restrictions directly; refusing to run unsandboxed

Removing `projects.<agent-home>.trust_level` did not address this second failure and also removed a valid defense-in-depth setting.

This PR restores the documented project trust override and selects `windows.sandbox="elevated"` on native Windows, allowing Cumora's existing split filesystem permission profile to be enforced without falling back to unsandboxed execution.

After rebuilding with this change:

- Codex big-brain and small-brain doctor probes both passed.
- The complete secure configuration loaded successfully.
- The local daemon started normally and hosted the Codex agents without the previous configuration or sandbox startup failures.
@yetone

bingqilinweimaotai:
BTW, in the future, PRs involving major changes may need to undergo rigorous validation before they can be merged. Merging them hastily could break some previously working features. For instance, the previous security BYOA issue caused Codex to be unavailable in the current version.

yetone:
Thanks for chasing this, and the Windows half of your analysis may well be right — but I can't merge it as it stands, because production contradicts the claim the rest of it rests on.

## The dotted `projects.*` form is rejected on POSIX, on the versions you validated

> The `projects.<path>.trust_level` field is valid and strictly parses across the supported Codex versions tested.
> strict config parsing verified with Codex 0.138.0 and 0.151.0 · real `--doctor` verification with Codex 0.152.0

Over the last 30 hours, `unknown configuration field \`projects."…"\`` fired **252,589 times on POSIX** across **121 computers** — against 1,743 on Windows. Broken down by the codex the daemon actually reported from each machine:

| codex | POSIX rejections | computers |
|---|---:|---:|
| 0.147.0 | 38,567 | 13 |
| **0.151.0** | **35,721** | **18** |
| 0.149.1 | 20,070 | 6 |
| 0.139.0 | 16,143 | 4 |
| 0.145.0 | 14,492 | 5 |
| 0.150.1 | 13,428 | 5 |
| **0.152.0** | **13,153** | **12** |

0.151.0 and 0.152.0 are the two releases your validation covers. Eighteen and twelve machines respectively are refusing that override right now.

The POSIX/Windows split also settles the mechanism. `resolveSpawn` returns `shell:false` on every platform except Windows, so on POSIX there is no shell and nothing to strip quotes — which rules out the cmd.exe cause your 6a70087 correctly fixed, and leaves the spelling itself. That is why restoring the dotted form would not be defence-in-depth restored; it would be 121 computers back to failing every wake before the model starts.

I should say plainly that I made the same mistake in the other direction: my first pass at this reasoned from my own codex install and reached a conclusion that production then contradicted. A local `--doctor` pass tells us the flag parses on *that* machine and config; it cannot tell us what the field set is on someone else's.

## What already landed

#151 (@WhichPaths) solved this without giving anything up: same override, spelled as an inline table.

```
-c projects={"<agent home>"={trust_level="untrusted"}}
```

That is merged as of c2e48e5. It keeps `trust_level="untrusted"` — the boundary you're right to want to preserve, and which my earlier removal did give up — while avoiding the dotted dynamic key that codex refuses.

## What I'd like from this PR

The `windows.sandbox="elevated"` half is a separate claim I have no production signal on either way, and it is plausible: if codex's unelevated restricted-token sandbox genuinely cannot enforce a split filesystem profile, that matters and nobody else has looked at it. Please rebase onto main and resubmit **just that part** — the Windows sandbox selection and its test — on top of #151's inline form. I'll review it on its own merits.

Please also drop the test change that re-pins the exact dotted string. #151 rewrote that assertion to pin the intent (this home is named, and it is untrusted) precisely so the spelling can change again when a CLI forces it; pinning the literal is what made the last necessary rewrite look like a regression.

yetone:
Thanks for trimming this down to exactly the Windows half — this is what I asked for, and it reviews cleanly.

What I checked before merging:

- **The key is real and strict-parses on the floor we support.** `--strict-config -c 'windows.sandbox="elevated"'` is accepted by codex 0.138.0, 0.140.0 and 0.144.6 (a misspelled `windows.sandboxx` is rejected by the same probe, so the check is discriminating). It is also in the official config reference as `windows.sandbox = unelevated | elevated`. So no Windows daemon on a supported codex will start refusing the profile because of this line.
- **Blast radius is native Windows only.** The arg is behind `IS_WIN`; POSIX argv is byte-identical to main, and that is where the 121-computer incident lived. The test change is gated the same way.
- **Worst case is no worse than today.** Codex's docs say a machine where the elevated one-time setup was never approved falls back to `unelevated` — which is the state that currently refuses the split profile. So machines without the setup stay exactly where they are; machines with it get a sandboxed codex instead of nothing.
- Merges cleanly onto main on top of #151's inline `projects={…}` form (it does not touch that line); `typecheck`, `server:typecheck`, lint, all three guards and the engine test file pass on the merged tree.

One thing worth knowing for operators, which your doc line hints at but does not say: `elevated` depends on a one-time administrator-approved setup (UAC, local sandbox users, firewall rules). If a headless daemon on a fresh box hits that path without the setup done, the fallback is silent. If you see that in practice, a `--doctor` line saying "elevated sandbox not set up" would be a good follow-up — not a blocker here.

Merging. Ships in v0.13.1.

yetone:
Correction on the version: the v0.13.1 rollout was blocked by an unrelated migration sentinel bug (#161), so this shipped to production in **v0.13.2**.

PR #163, opened 2026-09-02
fix(byoa): complete engine registry wiring and default model pins for gemini and qwen

wg2038:
### Summary

When Bring-Your-Own-Agent (BYOA) engines `gemini` and `qwen` were added, several downstream mapping and fallback sites were not updated:

1. **`daemon.ts` auth failure hints**: `authFailureHint()` lacked branches for `gemini` and `qwen`, falling through to `return 'Open Codex on that computer and refresh its login or quota, then wake the agent again.'`.
2. **`daemon.ts` triage model resolution**: `triageModel()` lacked branches for `gemini` and `qwen`, falling through to `<cursor-default>` in the LLM cost ledger (`llm_calls`) rather than using their respective triage models (`gemini-2.5-flash-lite` and `qwen3-coder-flash`).
3. **`registry.ts` default model pins**: `listAgentsForComputer()` did not handle `CUMORA_DEFAULT_GEMINI_MODEL` and `CUMORA_DEFAULT_QWEN_MODEL`, causing agents without explicit model overrides to receive `null` instead of the deploy-level default.
4. **`registry.ts` type safety**: `ENGINE_BINS` was declared as `Record<string, string>`. It is now typed as `Record<Exclude<EngineId, 'managed'>, string>` (consistent with `PAIRABLE`) so missing engines are caught at compile time.
5. **`guard-engine-registry.mjs`**: Extended to assert `authFailureHint`, `triageModel`, `listAgentsForComputer`, and `ENGINE_BINS` across all engines in `ENGINE_IDS`.
6. **Documentation & Env Example**: Updated `docs/COORDINATION.md` and `.env.example` with `CUMORA_DEFAULT_PI_MODEL`, `CUMORA_DEFAULT_GEMINI_MODEL`, `CUMORA_DEFAULT_QWEN_MODEL`, and `CUMORA_DEFAULT_ANTIGRAVITY_MODEL`.

### Testing

- `npm run lint`: 452 files checked, 0 errors, 0 warnings.
- `npm run typecheck` & `npm run server:typecheck`: 0 errors.
- `npm run guard:big-brain`, `guard:llm-tracked`, `guard:engine-registry`: all passed.
- `npm test`: 1058 passing (added `server/src/__tests__/agents-computer-daemon-auth-hint.test.ts` and updated `agents-computer-engine-detect.test.ts`).
- `npm run test:integration`: 269 passing.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,19 @@ OPENAI_API_KEY=sk-...
 # OPENAI_COMPACTION_MODEL=gpt-5.4-mini
 # OPENAI_IMAGE_MODEL=gpt-image-2
 
+# ─── BYOA default model pins (optional) ──────────────────────────────────
+# Deploy-level default models for paired BYOA computers when an agent row has
+# no explicit model override. See docs/COORDINATION.md.
+# CUMORA_DEFAULT_CLAUDE_MODEL=claude-opus-4-7
+# CUMORA_DEFAULT_CODEX_MODEL=
+# CUMORA_DEFAULT_GROK_MODEL=
+# CUMORA_DEFAULT_CURSOR_MODEL=
+# CUMORA_DEFAULT_OPENCODE_MODEL=
+# CUMORA_DEFAULT_PI_MODEL=
+# CUMORA_DEFAULT_GEMINI_MODEL=
+# CUMORA_DEFAULT_QWEN_MODEL=
+# CUMORA_DEFAULT_ANTIGRAVITY_MODEL=
+
 # ─── OrcaRouter (optional) ─────────────────────────────────────────────────
 # OpenAI-compatible AI gateway (https://www.orcarouter.ai). An agent opts in
 # by prefixing its model id with `orcarouter/`, e.g. `orcarouter/openai/gpt-4o-mini`.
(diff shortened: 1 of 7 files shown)


WhichPaths:
I wrote the gemini and qwen adapters (#99, #113) and the registry guard (#114), so these are my loose ends — thanks for running them down. I checked out the branch and exercised the guard rather than reading it, and the new checks do what they claim:

```
# deleted the gemini branch from authFailureHint
server/src/agents/computer/daemon.ts → authFailureHint
  'gemini' is missing — add an auth failure hint for 'gemini' in authFailureHint

# deleted the qwen branch from triageModel
server/src/agents/computer/daemon.ts → triageModel
  'qwen' is missing — add triage model resolution for 'qwen' in triageModel
```

I also checked the failure mode of the four new anchors, since a source-scraping check that silently matches nothing is worse than no check. They fail closed — rewriting `return rows.map((r) => {` into `const mapped = rows.map((r) => {` gives:

```
server/src/agents/computer/registry.ts
  anchor not found: listAgentsForComputer default model fallback — update this guard alongside the refactor
```

Two things I'd change, both about the fallbacks rather than the wiring:

**`triageModel`'s fallback still labels unknown engines as cursor.** With the new `cursor` branch added above it, all nine ids in `ENGINE_IDS` now return explicitly, so `return this.agent.model ?? '<cursor-default>'` is reachable only for an engine that is *not* in `ENGINE_IDS` — and it writes that engine's cost rows into `llm_calls` under cursor's name. That is the same mislabelling this PR is fixing, just one engine later. The guard makes the branch mandatory, so nothing legitimate lands there any more; `'<unknown-engine-default>'` would keep the row honest.

**`authFailureHint`'s new default drops the only actionable sentence.** The early return above it already covers everything that is *not* auth-shaped, so the final `return` is reached only when the detail **is** auth/quota/billing-shaped and no engine matched. Sending that case to "Check the daemon terminal for details" is strictly less than it had: the operator now learns nothing about the auth failure the regex just confirmed, and the daemon terminal is usually on a machine they are not sitting at — the argument behind #138. Something engine-agnostic but still actionable keeps it, e.g.:

```ts
return 'Sign in to that engine\'s CLI on that computer (or fix its API key / quota), then wake the agent again.'
```

Neither blocks the fix; the wiring itself and the guard coverage look right to me.

PR #177, opened 2026-09-03
fix(antigravity): resolve model catalog discovery and pairing engine assignment

wg2038:
## Summary

This PR addresses two related gaps affecting Antigravity engine support in Cumora:

1. **Antigravity Model Catalog Discovery & Small-Brain Triage Selection**:
   - In PR #175 (`feat(byoa): add account-specific model selection`), Antigravity was defined with an empty preset array (`models: []`), no discovery handler in `discoverEngineModelCatalog`, and `fastModelScope: 'computer'`. This resulted in an empty model dropdown in the UI (only "Follow engine default") and hid the fast-model triage picker.
   - Discovers dynamic model catalog via the built-in `agy models` CLI command, parsing tab-separated ID and human-readable label.
   - Categorizes models into recommended big/small tiers (using boundary-delimited matching to avoid false positives like `gemini` matching `mini`).
   - Adds standard fallback presets for `antigravity` and `qwen`, and configures `fastModelScope: 'agent'` so agents on Antigravity/Qwen hosts can configure their triage model.
   - Provides a default triage model fallback in `daemon.ts`.

2. **Fix Default Engine on Computer Pairing**:
   - In `server/src/api/router.ts` (`POST /computers/pair`), a hardcoded disjunction enumerated pairable engines for free-tier auto-onboarding but omitted `'antigravity'`.
   - Pairing a machine with `--engine antigravity` caused `engine` to incorrectly fall back to `'claude'`, inappropriately setting stranded and newly seeded starter agents to `'claude'` on machines without Claude installed.
   - Replaces the hardcoded check with the authoritative `PAIRABLE_ENGINES` set exported from `registry.ts`.

## Verification

- `npm run lint`: Checked 478 files, 0 errors, 0 warnings.
- `npm run typecheck`: Passed (0 errors).
- `npm run server:typecheck`: Passed (0 errors).
- `npm run guard:engine-registry`: Passed.
- `npm run guard:big-brain`: Passed.
- `npm run guard:llm-tracked`: Passed.
- Full test suite: **1106 passed**, 0 failed, 4 skipped.

### DIFF
--- a/server/src/__tests__/agent-computer-pairing-token.test.ts
+++ b/server/src/__tests__/agent-computer-pairing-token.test.ts
@@ -149,3 +149,29 @@ test('daemon run mode (supervised) is persisted on pair and heartbeat', async ()
   assert.equal(hb.params[1], '0.1.122')
   assert.equal(hb.params[2], false)
 })
+
+test('pairComputer accepts antigravity engine as primary and stores it in available_engines', async () => {
+  const calls = installPoolMock(({ sql }) => {
+    if (/SELECT pair_token FROM companies/.test(sql)) return { rows: [{ pair_token: 'company-token' }] }
+    if (/SELECT id, company_id/.test(sql)) return { rows: [] }
+    if (/SELECT id AS company_id, owner_user_id FROM companies/.test(sql)) {
+      return { rows: [{ company_id: 'co-1', owner_user_id: 'u-1' }] }
+    }
+    if (/SELECT id FROM computers/.test(sql)) return { rows: [{ id: 'comp-antigravity' }] }
+    if (/UPDATE computers\s+SET credential_hash/.test(sql)) return { rowCount: 1 }
+    throw new Error(`unexpected query: ${sql}`)
+  })
+
+  const paired = await registry.pairComputer({
+    code: 'company-token',
+    hostName: 'Dev Box',
+    engines: ['antigravity', 'claude'],
+    deferBroadcast: true,
+  })
+  assert.equal(paired?.computerId, 'comp-antigravity')
+
+  const update = calls.find((c) => /UPDATE computers\s+SET credential_hash/.test(c.sql))
+  assert.ok(update)
+  assert.equal(update.params[1], JSON.stringify(['antigravity', 'claude']))
+})
+
--- a/server/src/__tests__/agents-computer-model-catalog.test.ts
+++ b/server/src/__tests__/agents-computer-model-catalog.test.ts
@@ -93,3 +93,19 @@ test('Claude settings ignore a relative config-root override instead of reading 
     prefersLocalDefault: false,
   })
 })
+
+test('Antigravity catalog parses tab-separated models, ignores fetching banner, and marks recommendations', () => {
+  const output = [
+    'Fetching available models...',
+    'gemini-3.8-flash-high\tGemini 3.8 Flash (High)',
+    'gemini-3.1-pro-high\tGemini 3.1 Pro (High)',
+    'claude-sonnet-4-6\tClaude Sonnet 4.6 (Thinking)',
+  ].join('\n')
+
+  const out = parseListedModels(output, 'antigravity')
+  assert.deepEqual(out, [
+    { id: 'gemini-3.8-flash-high', label: 'Gemini 3.8 Flash (High)', description: null, recommendedFor: ['small'] },
+    { id: 'gemini-3.1-pro-high', label: 'Gemini 3.1 Pro (High)', description: null, recommendedFor: ['big'] },
+    { id: 'claude-sonnet-4-6', label: 'Claude Sonnet 4.6 (Thinking)', description: null, recommendedFor: ['big'] },
+  ])
+})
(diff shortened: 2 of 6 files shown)


WhichPaths:
The pairing half of this is the right fix, and I want to put a number on why — I wrote the `PAIRABLE` record (#116) and the engine-registry guard (#114), and the guard did **not** catch this.

On current `main`, with the bug present:

```
$ grep -c antigravity server/src/api/router.ts
0
$ node scripts/guard-engine-registry.mjs
✅ engine registry guard: every engine in ENGINE_IDS is wired into all of its lists.
```

Green, with a pairing route that silently downgrades `--engine antigravity` to `claude`. That is worth stating plainly so nobody reads that check mark as covering this class: the guard walks **named lists** it has anchors for (`ENGINE_BINS`, `PAIRABLE`, `ENGINE_VERSION_SPECS`, `BYOA_SOURCES`, the source unions, R4). A membership test written as a chain of `===` against string literals is not a list, so neither the guard nor the compiler could relate it to `EngineId` — the disjunction is well-typed whichever engines it happens to name.

Replacing it with `PAIRABLE_ENGINES.has(...)` is the fix that actually holds, because that set derives from

```ts
const PAIRABLE: Record<Exclude<EngineId, 'managed'>, true> = { … }
```

which is compiler-enforced: widening `EngineId` fails the build until `PAIRABLE` gains the key, and the set follows for free. The next engine cannot repeat this.

I swept for other instances of the same shape so you know whether this was the last one — `===`, `.includes('<engine>')`, and `switch`/`case` on engine literals across `server/src`, `src`, `electron`, and `agent-cli`, excluding tests and migrations. The only multi-engine literal enumerations left are the deliberate named ones: `ENGINE_IDS`, `SANDBOXED_ENGINE_IDS` (a subset on purpose), `src/lib/engines.ts`'s `RUNNABLE_ENGINES`, and the per-engine `if (engine === 'x')` branch chains the guard already covers by name. **This route was the only free-floating one.**

One stale leftover, unrelated to your change and only a comment — `server/src/db/migrate.ts:1568` documents the column as

```
available_engines JSONB NOT NULL DEFAULT '[]'::jsonb,    -- ['claude','codex','grok','cursor','opencode','pi','gemini','qwen']; ['managed'] for cloud
```

which predates both `qwen`'s successor and `antigravity`. Harmless, but it is the kind of list a future reader copies from.

The catalog half reads well to me too — the boundary-delimited tier matching is the right call, `gemini` matching `mini` under a naive substring test is exactly the sort of thing that would have mislabelled a big model as the cheap triage one.

yetone:
Reviewed — both halves of this are correct and I want them.

The pairing fix in particular is the kind of bug that is invisible until someone hits it: `POST /computers/pair` enumerating pairable engines by hand meant every engine added after that line was written silently fell back to `'claude'`, seeding starter agents onto an engine the machine doesn't even have installed. Replacing it with the authoritative `PAIRABLE_ENGINES` set from `registry.ts` is the right fix and it fixes it for the *next* engine too, not just antigravity.

I can't merge it right now: GitHub reports the branch as `CONFLICTING` / `DIRTY`. I merged 16 PRs into `main` today and several of them touch `server/src/api/router.ts` and `server/src/agents/computer/daemon.ts`, so the conflict is almost certainly mine, not yours. Sorry about that.

Please rebase onto current `main` and I'll merge it. Two things to re-check while you're in there:

- `model-catalog.ts` `PRESETS` is a `Record<EngineId, EngineModelCatalog>`, and `main` has moved — make sure the qwen/antigravity preset edits still apply where you expect.
- The `daemon.ts` change to the antigravity default (`'<antigravity-default>'` → `'gemini-3.8-flash-high'`) pins a specific model id as the fallback. That's fine as a triage default, but please confirm it still resolves against a current `agy models` listing after the rebase — a hardcoded id that a provider retires becomes a hard failure on wake, where `<antigravity-default>` merely deferred.

Ping me when it's green and I'll take it.

wg2038:
@yetone Rebased cleanly onto `main` (on top of v0.15.0), conflicts resolved, and all CI checks are green!

Regarding the two items to re-check:

1. **`model-catalog.ts` PRESETS**:
   Confirmed that `PRESETS` (`Record<EngineId, EngineModelCatalog>`) cleanly retains the full model catalogue and agent-scoped triage configuration for both `qwen` (`qwen3-coder-flash`, `fastModelScope: 'agent'`) and `antigravity` (`gemini-3.8-flash-high`, `fastModelScope: 'agent'`).

2. **Antigravity Fallback Triage Model Resolution (`gemini-3.8-flash-high`)**:
   Confirmed against a live `agy models` invocation on the current Antigravity CLI build. The command outputs:
   ```text
   Fetching available models...
   gemini-3.8-flash-high	Gemini 3.8 Flash (High)
   gemini-3.8-flash-medium	Gemini 3.8 Flash (Medium)
   gemini-3.8-flash-low	Gemini 3.8 Flash (Low)
   gemini-3.7-flash-high	Gemini 3.7 Flash (High)
   gemini-3.1-pro-high	Gemini 3.1 Pro (High)
   claude-sonnet-4-6	Claude Sonnet 4.6 (Thinking)
   claude-opus-4-6-thinking	Claude Opus 4.6 (Thinking)
   gpt-oss-120b-medium	GPT-OSS 120B (Medium)
   ```
   `gemini-3.8-flash-high` is actively returned at the top of the catalogue as the primary flash tier, and the leading `Fetching available models...` line is cleanly skipped by `parseListedModels`.

Ready for merge! Thank you!

PR #183, opened 2026-09-04
fix(calendar): eliminate monthly and leap-year recurrence drift in frontend views

### LINKED ISSUE #182: fix(calendar): recurrence drift on month-end and leap years in desktop/mobile calendar views
## Problem
When recurring calendar events are scheduled on month-end dates (e.g. 31st), leap years (Feb 29), or with bi-weekly intervals (`interval: 2, byweekday: [1, 3]`), the client calendar views in `CalendarView.tsx` (desktop) and `MobileCalendar.tsx` (mobile) drift and diverge from the server dispatcher (`server/src/calendar.ts`).

For example:
1. **Monthly on Jan 31st**:
   - Backend fires: `01-31, 02-28, 03-31, 04-30, 05-31...`
   - Frontend displays: `01-31, 03-03, 04-03, 05-03...` (February is completely skipped, and all future months permanently drift to the 3rd).
2. **Yearly on Feb 29th**:
   - Backend fires: `2028-02-29, 2029-02-28, 2030-02-28, 2031-02-28, 2032-02-29...`
   - Frontend displays: `2028-02-29, 2029-03-01, 2030-03-01...` (Permanently drifted to March 1st).
3. **Bi-weekly on Mon & Wed**:
   - Backend fires: Mon and Wed in the active week, then skips the next week.
   - Frontend displays: Broken 9-day / 12-day / 9-day intervals instead of alternating bi-weekly pairs.

## Root Cause
In commit `21ea326` (#157), server-side recurrence calculation (`server/src/recurrence.ts`) was rewritten to calculate occurrences anchored on the `seed` with indexed arithmetic and date clamping.

However, client-side calendar renderers in `src/desktop/CalendarView.tsx` and `src/mobile/MobileCalendar.tsx` still duplicated the old `stepRule` logic with mutable stepping (`setMonth(getMonth() + interval)`), where month-end overflows in JavaScript `Date` permanently displace all subsequent occurrences.

## Proposed Solution
1. Extract pure client-side recurrence calculation into `src/lib/recurrence.ts` (`nextOccurrenceOnOrAfter`), matching the indexed math in `server/src/recurrence.ts`.
2. Replace duplicate `stepRule` implementations in `CalendarView.tsx` and `MobileCalendar.tsx` with `nextOccurrenceOnOrAfter`.
3. Add unit tests ensuring client recurrence calculation produces identical occurrences to the server dispatcher across month-end, leap year, and bi-weekly weekday rules.

hasak21:
Fixes #182

## Summary
In commit `21ea326` (#157), server-side recurrence calculation (`server/src/recurrence.ts`) was rewritten to calculate occurrences anchored on the `seed` with indexed arithmetic and clamping, preventing date drift on month-end, leap year, and bi-weekly weekday rules.

However, `src/desktop/CalendarView.tsx` and `src/mobile/MobileCalendar.tsx` still duplicated the old mutable `stepRule` logic (`setMonth(getMonth() + interval)`), causing client-side agenda views to diverge from the backend dispatcher:
- **Monthly on Jan 31**: Client displayed `01-31, 03-03, 04-03...` (February skipped, permanent drift to 3rd) while server fired `01-31, 02-28, 03-31...`.
- **Yearly on Feb 29**: Client displayed `2028-02-29, 2029-03-01...` (permanent drift to March 1) while server fired `2028-02-29, 2029-02-28... 2032-02-29`.
- **Bi-weekly Mon/Wed**: Client produced broken 9-day/12-day/9-day intervals instead of alternating bi-weekly pairs.

## Changes
- Created `src/lib/recurrence.ts` (`nextOccurrenceOnOrAfter`, `occurrenceAt`, `addMonthsClamped`, `weeklyDays`, `addDays`) mirroring the indexed date math from `server/src/recurrence.ts`.
- Simplified `nextOccurrence` in both `src/desktop/CalendarView.tsx` and `src/mobile/MobileCalendar.tsx` to delegate to `nextOccurrenceOnOrAfter`.
- Added unit tests in `server/src/__tests__/frontend-calendar-recurrence.test.ts` validating parity across all edge cases.

## Verification
- `npm run lint`: passed (checked 482 files, no issues)
- `npm run typecheck`: passed
- `npm run server:typecheck`: passed
- `npm run guard:big-brain && npm run guard:llm-tracked && npm run guard:engine-registry`: passed
- `npm test`: passed (1114 passing, 0 failing)
- `npm run build`: passed (Vite production bundle built successfully)

### DIFF
--- a/server/src/__tests__/frontend-calendar-recurrence.test.ts
+++ b/server/src/__tests__/frontend-calendar-recurrence.test.ts
@@ -51,3 +51,10 @@ test('frontend recurrence: every-2-weeks on Mon and Wed keeps both days in the s
     ],
   )
 })
+
+test('frontend recurrence: non-recurring event fires once then terminates', () => {
+  const seed = new Date('2026-04-14T09:00:00Z')
+  assert.deepEqual(nextOccurrenceOnOrAfter(seed, null, seed), seed)
+  assert.equal(nextOccurrenceOnOrAfter(seed, null, new Date(seed.getTime() + 1000)), null)
+})
+
--- a/server/src/__tests__/frontend-calendar-recurrence.test.ts
+++ b/server/src/__tests__/frontend-calendar-recurrence.test.ts
@@ -1,0 +1,53 @@
+/**
+ * Test frontend recurrence math against the exact month-end, leap year, and
+ * bi-weekly test cases to ensure client calendar views stay in parity with the
+ * backend dispatcher.
+ */
+import { test } from 'node:test'
+import assert from 'node:assert/strict'
+import { nextOccurrenceOnOrAfter } from '../../../src/lib/recurrence.js'
+import type { RecurrenceRule } from '../../../src/types.js'
+
+const DOW = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat']
+
+function series(seedIso: string, rule: RecurrenceRule | null, n: number): string[] {
+  const seed = new Date(seedIso)
+  const out: string[] = []
+  let after = seed
+  for (let i = 0; i < n; i++) {
+    const next = nextOccurrenceOnOrAfter(seed, rule, after)
+    if (!next) break
+    out.push(next.toISOString().slice(0, 10))
+    after = new Date(next.getTime() + 1000)
+  }
+  return out
+}
+
+function weekdays(seedIso: string, rule: RecurrenceRule, n: number): string[] {
+  return series(seedIso, rule, n).map((d) => `${d} ${DOW[new Date(`${d}T00:00:00Z`).getUTCDay()]}`)
+}
+
+test('frontend recurrence: monthly on the 31st clamps into short months and comes back', () => {
+  assert.deepEqual(
+    series('2026-01-31T09:00:00Z', { freq: 'monthly', interval: 1 }, 7),
+    ['2026-01-31', '2026-02-28', '2026-03-31', '2026-04-30', '2026-05-31', '2026-06-30', '2026-07-31'],
+  )
+})
+
+test('frontend recurrence: yearly on Feb 29 returns at the next leap year', () => {
+  assert.deepEqual(
+    series('2028-02-29T09:00:00Z', { freq: 'yearly', interval: 1 }, 6),
+    ['2028-02-29', '2029-02-28', '2030-02-28', '2031-02-28', '2032-02-29', '2033-02-28'],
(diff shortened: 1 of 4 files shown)


WhichPaths:
I wrote the server-side fix this mirrors (#157), so I checked the new copy the way I'd want mine checked: I ran **all 17 cases** from `server/src/__tests__/calendar-recurrence.test.ts` against `src/lib/recurrence.ts` and diffed the two implementations occurrence-by-occurrence.

**Zero divergence.** The 13 shapes your test file does not cover — monthly on the 30th, February clamping under `interval: 2` across a leap year, mid-month monthly, weekly `interval: 1` on Mon+Wed, the seed-week no-backfill rule, weekly with no `byweekday`, a duplicated/unsorted `byweekday`, an empty `byweekday`, daily with an interval, `count` including the seed, `until` inclusive, `until` terminating a clamped monthly series, and the far-future skip-ahead — all agree with the server exactly. The port is faithful.

One structural note, offered as a follow-up rather than a change request.

This PR resolves the drift by adding a **second** implementation and a **second** hand-written expectation table. That is the same shape that produced #182: `server/src/recurrence.ts` was fixed in #157, and nothing told the frontend copy about it for a release. After this lands there are two implementations and two tables, and the next change to the server's math has the same silent path back — update `calendar-recurrence.test.ts`, leave `frontend-calendar-recurrence.test.ts` as it is, and the two diverge again with CI green.

The asymmetry is already visible: 17 cases on one side, 4 on the other, transcribed by hand.

What would actually pin it is a **differential** test — run both implementations over one shared case table and assert they produce identical series, instead of asserting each against its own copy of the answers:

```ts
for (const c of CASES) {
  assert.deepEqual(series(serverImpl, c), series(frontendImpl, c), c.name)
}
```

That is what I ran to write this comment, and it needs no expected-value table at all — divergence is the failure, whichever side moved. It also survives a future change to the math: update the server, and the frontend test goes red instead of going quietly stale.

Worth noting your test file already proves the mechanism works — importing `../../../src/lib/recurrence.js` from `server/src/__tests__/` is what lets `npm test` reach frontend code at all, since the renderer has no runner of its own. A parity test rides the same trick.

I'm happy to send that as a follow-up once this merges — it is additive and shouldn't hold up the fix, which is correct as it stands.

hasak21:
Thanks for the thorough review and verification, @WhichPaths! Fully agree on the differential parity test idea — looking forward to the follow-up once this lands!

PR #230, opened 2026-09-06
fix(byoa): isolate sessions per engine and retry once fresh on stale resume (#209)

### LINKED ISSUE #209: BYOA engine hot-switch reuses Codex session IDs in Claude, causing an infinite resume failure loop
## Summary

Cumora 0.14.2 stores one session ID per agent without recording which engine created it. If the daemon temporarily falls back from Claude Code to Codex and later switches back to Claude, it passes the Codex session ID to `claude --resume`.

Claude rejects the ID with `No conversation found with session ID`, but the daemon keeps retrying the same ID indefinitely instead of starting a fresh Claude session.

## Environment

- Cumora CLI: 0.14.2
- OS: macOS
- Service mode: LaunchAgent
- Engines installed: Claude Code and Codex
- Preferred agent engine: Claude
- Claude secure minimum: 2.1.248
- Codex secure minimum: 0.138.0

## Observed sequence

At daemon startup, Claude version verification temporarily failed:

```text
secure engine(s) disabled: claude (could not verify the installed version (secure minimum 2.1.248))
cumora 0.14.2 ... (engines: codex)
```

Cumora then hosted the affected agents on Codex. Codex could not restore their previous sessions and created fresh Codex sessions.

Shortly afterward, Claude became detectable again:

```text
engines on PATH changed: codex → claude, codex
agent Atlas (atlas-a745) config changed → restarting on claude
atlas-a745 restored engine session 01a06f71... from disk — will --resume
```

Cumora then passed the newly saved Codex session ID to Claude:

```text
atlas-a745 engine session respawned (resume 01a06f71)
No conversation found with session ID: 01a06f71-...
[session] engine process died while idle: exited with code 1
```

The daemon repeated this sequence continuously for multiple agents.

I verified that all four rejected session IDs existed under:

```text
~/.codex/archived_sessions/
```

and did not exist under:

```text
~/.claude/projects/
```

This confirms that Claude was being asked to resume Codex-owned session IDs.

## Expected behavior

A session ID should only be resumed by the engine that created it.

When an agent changes engines, Cumora should either:

1. start a fresh session for the destination engine; or
2. restore a session previously created by that same engine.

If an engine reports that a resume target does not exist, Cumora should clear the stale session ID and retry once with a fresh session.

## Actual behavior

The same per-agent `.session` file is reused across engines. A Codex session ID can therefore be passed to Claude, permanently wedging the agent in a retry loop.

Cumora 0.14.2 contains stale-session detection, but this path reduces the underlying error to:

```text
engine turn error (...): see log
```

The `No conversation found` text is therefore unavailable to the stale-resume matcher, so the saved ID is never cleared.

## Suggested fix

- Persist both the engine type and session ID:

```json
{
  "engine": "codex",
  "sessionId": "..."
}
```

- Alternatively, maintain separate session files for each engine.
- Never pass a saved session ID to a different engine.
- Start fresh when the destination engine has no compatible saved session.
- Preserve the underlying engine error when deciding whether a resume target is stale.
- Clear the saved ID and retry once when the engine explicitly reports that the session does not exist.
- Retry or debounce transient engine-version probe failures before switching all agents to another engine.

## Workaround

Stop the daemon, back up and remove the affected per-agent `.session` files, then restart Cumora so each agent creates a fresh session.

wg2038:
Closes #209

### Background & Root Cause

In Cumora BYOA, session IDs were stored in a single unversioned file per agent (`~/.cumora/sessions/<agentId>.session`).

When local engine availability drifted (e.g. transient version probe failure causing fallback from Claude Code to Codex, and subsequently recovering back to Claude):
1. The agent runner on Claude restored the session ID created by Codex;
2. `claude --resume <codex-session-id>` was invoked;
3. Claude rejected the ID with `No conversation found with session ID`;
4. `ClaudeSession.onStdout` masked the error event as `engine turn error (...): see log` because `ev.result` was not a string, losing the underlying failure diagnostic;
5. As a result, `isStaleResumeError()` failed to match `'see log'`, the dead session ID was never cleared from disk, and `persistentUnusable` was permanently set to `true`;
6. The daemon was trapped in an infinite retry failure loop across wakes.

### Changes

1. **Engine Session File Isolation**:
   - Scoped session files to each engine: `sessionFileFor(agentId, engine)` maps to `~/.cumora/sessions/<agentId>.<engine>.session`.
   - Different engines maintain independent continuity without ever cross-resuming.
   - On startup, legacy unversioned `<agentId>.session` files are safely purged rather than adopted into unverified engines.

2. **Error Preservation in `ClaudeSession`**:
   - In `ClaudeSession.onStdout`, when an error `result` event is processed, error details are extracted from `ev.result`, `ev.error`, `ev.error.message`, and the tail of `stderr` / unparsed `stdout` lines before falling back to `'see log'`.
   - This ensures `No conversation found with session ID` is preserved for `isStaleResumeError()`.

3. **Stale Resume Single Fresh Retry**:
   - Extracted `runEngineAttempt` shared by chat and agenda turns.
   - Stale resume errors no longer poison `persistentUnusable`.
   - When an engine explicitly reports a stale resume target, the stale session file is cleared and the turn is retried **once fresh** (`resumeSessionId: null`), succeeding in the same wake.

### Verification

- New unit tests in `server/src/__tests__/agents-computer-session-isolation.test.ts`:
  - `sessionFileFor` engine scoping;
  - Multi-engine disk state isolation across hot-switches;
  - Legacy session purge;
  - Detection of unmasked Claude resume errors;
  - Single-retry fresh recovery behavior.
- Existing tests:
  - `agents-computer-session-reset.test.ts`: 15/15 passed
  - All 32 `agents-computer-*.test.ts`: 249 passed, 0 failed
- Quality checks:
  - `npm run lint` (Biome): clean, 0 errors
  - `npm run typecheck` & `npm run server:typecheck`: 0 errors
  - Source guards (`guard:big-brain`, `guard:llm-tracked`, `guard:engine-registry`): all passed

### DIFF
--- a/server/src/__tests__/agents-computer-session-isolation.test.ts
+++ b/server/src/__tests__/agents-computer-session-isolation.test.ts
@@ -1,0 +1,354 @@
+/**
+ * Tests for Issue #209:
+ * BYOA engine hot-switch reuses Codex session IDs in Claude, causing an infinite resume failure loop.
+ *
+ * Verifies:
+ * 1. Session files are strictly scoped per engine: `<agentId>.<engine>.session`
+ * 2. AgentRunner methods loadSessionId/persistSessionId isolate session state across engine switches
+ * 3. AgentRunner.loadSessionId safely purges legacy unscoped session files (`<agentId>.session`)
+ * 4. isStaleResumeError accurately detects unmasked Claude resume errors
+ * 5. AgentRunner.mustResetSession distinguishes stale resumes from fresh runs and context overflows
+ * 6. AgentRunner.runEngineTurn retries once with resumeSessionId: null on stale-resume error and succeeds
+ * 7. AgentRunner.runEngineTurn never retries a turn that reported model usage (preventing duplicate side effects)
+ *
+ * Run: node --import tsx --test server/src/__tests__/agents-computer-session-isolation.test.ts
+ */
+import { test } from 'node:test'
+import assert from 'node:assert/strict'
+import { mkdir, readFile, rm, writeFile } from 'node:fs/promises'
+import { join } from 'node:path'
+import {
+  AgentRunner,
+  type AgentInfo,
+  type DaemonConfig,
+  hasNonZeroUsage,
+  isStaleResumeError,
+  sessionFileFor,
+  SESSIONS_DIR,
+} from '../agents/computer/daemon.js'
+import type { EngineRunResult } from '../agents/computer/engine.js'
+
+function makeConfig(): DaemonConfig {
+  return {
+    serverUrl: 'http://127.0.0.1:8787',
+    computerId: 'test-comp',
+    deviceToken: 'test-device-token',
+  }
+}
+
+function makeAgent(id: string): AgentInfo {
+  return {
+    id,
+    name: `Agent ${id}`,
+    role: 'tester',
+    systemPrompt: 'Standing instructions',
+    engine: 'claude',
+    model: null,
+    fastModel: null,
+  }
+}
+
+function uniqueAgentId(prefix: string): string {
+  return `${prefix}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`
+}
+
+test('sessionFileFor scopes the session path to the specific engine', () => {
+  const agentId = 'atlas-a745'
+  const claudeFile = sessionFileFor(agentId, 'claude')
+  const codexFile = sessionFileFor(agentId, 'codex')
+  const geminiFile = sessionFileFor(agentId, 'gemini')
+
+  assert.equal(claudeFile, join(SESSIONS_DIR, 'atlas-a745.claude.session'))
+  assert.equal(codexFile, join(SESSIONS_DIR, 'atlas-a745.codex.session'))
+  assert.equal(geminiFile, join(SESSIONS_DIR, 'atlas-a745.gemini.session'))
+  assert.notEqual(claudeFile, codexFile)
+  assert.notEqual(claudeFile, geminiFile)
+})
+
+test('switching engines isolates session state on disk via AgentRunner', async () => {
+  const agentId = uniqueAgentId('agent-switch')
+  const cfg = makeConfig()
+  const agent = makeAgent(agentId)
+
+  const claudeRunner = new AgentRunner(cfg, agent, 'claude')
+  const codexRunner = new AgentRunner(cfg, agent, 'codex')
+
+  try {
+    assert.equal(claudeRunner.sessionFile, join(SESSIONS_DIR, `${agentId}.claude.session`))
+    assert.equal(codexRunner.sessionFile, join(SESSIONS_DIR, `${agentId}.codex.session`))
+
+    // Agent starts on Claude and persists a session ID
+    claudeRunner.setSessionId('claude-uuid-1234')
+    await claudeRunner.persistSessionId()
+    assert.equal(claudeRunner.currentSessionId, 'claude-uuid-1234')
+    assert.equal(await readFile(claudeRunner.sessionFile, 'utf8'), 'claude-uuid-1234')
+
+    // Hot-switch to Codex: Codex loads its session file and sees nothing
+    assert.equal(codexRunner.currentSessionId, null)
+    await codexRunner.loadSessionId()
+    assert.equal(codexRunner.currentSessionId, null, 'Codex must start fresh, not inherit Claude session')
+
+    // Codex persists its own thread session
+    codexRunner.setSessionId('codex-thread-5678')
+    await codexRunner.persistSessionId()
+    assert.equal(codexRunner.currentSessionId, 'codex-thread-5678')
+    assert.equal(await readFile(codexRunner.sessionFile, 'utf8'), 'codex-thread-5678')
+
+    // Hot-switch back to Claude: a new Claude runner restores its own session untouched
+    const restoredClaude = new AgentRunner(cfg, agent, 'claude')
+    await restoredClaude.loadSessionId()
+    assert.equal(restoredClaude.currentSessionId, 'claude-uuid-1234', 'Claude restored its own prior session')
+    assert.notEqual(restoredClaude.currentSessionId, codexRunner.currentSessionId)
+  } finally {
+    await rm(claudeRunner.sessionFile, { force: true })
+    await rm(codexRunner.sessionFile, { force: true })
+  }
+})
+
+test('legacy unscoped session files are purged by AgentRunner.loadSessionId', async () => {
+  const agentId = uniqueAgentId('agent-legacy')
+  const cfg = makeConfig()
+  const agent = makeAgent(agentId)
+  const legacyFile = join(SESSIONS_DIR, `${agentId}.session`)
+  const runner = new AgentRunner(cfg, agent, 'claude')
+
+  try {
+    await mkdir(SESSIONS_DIR, { recursive: true })
+    // Legacy unversioned session written before session isolation (could be from any engine)
+    await writeFile(legacyFile, 'unverified-session-id', 'utf8')
+
+    // AgentRunner loads its engine-scoped session:
+    // It must purge the legacy file and start fresh
+    await runner.loadSessionId()
+
+    assert.equal(runner.currentSessionId, null, 'No session adopted from legacy unscoped file')
+
+    let legacyExists = true
+    try {
+      await readFile(legacyFile, 'utf8')
+    } catch {
+      legacyExists = false
+    }
+    assert.equal(legacyExists, false, 'Legacy unversioned session file was purged by loadSessionId')
+  } finally {
+    await rm(legacyFile, { force: true })
+    await rm(runner.sessionFile, { force: true })
+  }
+})
+
+test('isStaleResumeError accurately detects unmasked Claude resume errors', () => {
+  // When ClaudeSession extracts the actual error from ev.error / stderr:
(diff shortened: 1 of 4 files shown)


yetone (changes requested):
Thanks for this — the diagnosis in #209 is exactly right and three of the four changes here are ones I want:

- **engine-scoped session files** (`<agentId>.<engine>.session`) — correct, and refusing to adopt the legacy unscoped file rather than guessing its engine is the right call. It costs every existing agent one cold start on upgrade, which is the cheap side of that trade.
- **preserving the real error in `ClaudeSession.onStdout`** — `'see log'` swallowing `No conversation found with session ID` is the actual root cause of the wedge, and pulling it out of `ev.error` / `ev.error.message` / the stderr tail is the right fix.
- **not setting `persistentUnusable` on a stale resume** — a missing session is not this machine telling us the persistent transport is broken. Agreed.

I can't merge it as it stands, though, because of the fourth change.

### Blocking: the in-turn retry is scoped wider than stale-resume, and can duplicate a turn's side effects

Both call sites (`daemon.ts:2617` agenda, `daemon.ts:3007` chat) gate the fresh retry on:

```ts
if (result.error && resumeSessionId && this.mustResetSession(result.error, true)) {
```

but `mustResetSession` is true for three things, not one:

```ts
if (this.isContextOverflow(err)) return true
if (this.isPoisonedTranscript(err)) return true
if (hadResume && isStaleResumeError(err)) return true
```

For a stale resume the retry is unambiguously safe: the engine rejected the `--resume` target before running anything, so zero work happened and re-running cannot duplicate it. That is the case #209 is about.

Context-window overflow and a poisoned transcript are not that case. `CONTEXT_OVERFLOW_RE` matches `prompt is too long` / `input is too long`, which Claude Code raises on an API request that may be hop *N* of a multi-hop turn — after the agent has already posted messages, moved cards, or written calendar entries through the MCP bridge. The final `result` event still carries `usage`, and the persistent process is often still `alive`. Under this PR that turn now gets re-run from the same delta on a fresh session, and every side effect of hops 1..N−1 happens a second time. Double-posting into a conversation is the visible failure; double-charging the user's tokens is the quiet one.

This is the exact hazard the code immediately above it already guards against, and the guard was dropped on the new path:

```ts
// No usage means the model never ran — the process died in the
// transport, not mid-answer, so re-running the turn cannot duplicate
// work or double-charge.
if (result.error && !result.usage && !session.alive) {
```

It also silently contradicts the user-facing copy in `authFailureHint`, which still promises overflow is handled by *"session has been reset automatically — just wake the agent again"*.

**The fix I'd take:** scope the retry to the case you actually diagnosed —

```ts
if (result.error && resumeSessionId && isStaleResumeError(result.error)) {
```

and leave overflow/poison on the existing reset-and-recover-next-wake path (the second `mustResetSession(engineError, false)` call already handles them, since both are `hadResume`-independent). If you do want an in-turn retry for overflow later, it belongs in its own PR with the `!result.usage` guard carried over and the hint copy updated.

### Also, before I merge

The retry is the load-bearing new behaviour and has no test. `agents-computer-session-isolation.test.ts` covers `sessionFileFor` and `isStaleResumeError` well, but three of its five tests don't reach the code under test:

- *"switching engines isolates session state on disk"* hand-writes and hand-reads two paths; it never calls `loadSessionId`/`persistSessionId`, so it would still pass if `AgentRunner` ignored `sessionFileFor` entirely.
- *"legacy unscoped session files are purged"* calls `rm` itself and then asserts the file is gone — it asserts `node:fs` works, not that `loadSessionId` purges.
- *"mustResetSession behavior on stale resume vs fresh runs"* never calls `mustResetSession`; it's a duplicate of the `isStaleResumeError` test above it.

Please add one test that actually exercises `runEngineAttempt` retrying once with `resumeSessionId: null` after a stale-resume error and succeeding, and one that asserts a turn that *did* report usage is **not** retried. Those two are what would have caught the scoping problem.

Rebase on `main` when you push — `v0.16.2` has landed since you opened this.

wg2038:
Thanks for the thorough review and for catching the hazard on mid-turn side effects and double charging @yetone!

All requested changes have been implemented:

### 1. In-turn retry scoped strictly to stale resumes without usage
- Narrowed the retry trigger in `AgentRunner.runEngineTurn` to:
  ```ts
  if (result.error && !result.usage && resumeSessionId && isStaleResumeError(result.error)) {
    console.log(`[computer] ${this.agent.id} resume target is stale — resetting and retrying turn fresh`)
    await this.resetEngineSession('stale resume target')
    result = await this.runEngineAttempt(delta, null, purpose)
  }
  ```
- Context overflow and poisoned transcript remain exclusively on the existing post-turn reset path:
  ```ts
  if (engineError && this.mustResetSession(engineError, false)) {
    await this.resetEngineSession(this.resetReason(engineError))
  }
  ```
  This preserves the *"session has been reset automatically — just wake the agent again"* contract and completely eliminates any risk of repeating hops 1..N-1 side effects or double-billing tokens.
- Extracted into `runEngineTurn`, cleanly unifying the retry path for both chat turns (`runTurn`) and proactive board work (`runAgenda`).

### 2. Exported `AgentRunner` & refactored session isolation tests
- Exported `AgentRunner`, `DaemonConfig`, `AgentInfo`, and exposed `loadSessionId`, `persistSessionId`, `currentSessionId`, `mustResetSession`, `runEngineAttempt`, and `runEngineTurn`.
- Refactored `server/src/__tests__/agents-computer-session-isolation.test.ts` to directly exercise `AgentRunner` methods:
  - **Engine switching**: Directly calls `runner.setSessionId()`, `runner.persistSessionId()`, and `runner.loadSessionId()` across Claude and Codex runners to verify disk state isolation and continuity recovery.
  - **Legacy purge**: Calls `runner.loadSessionId()` and asserts the unscoped `<agentId>.session` file is purged from disk while starting the engine cold.
  - **mustResetSession**: Calls `runner.mustResetSession()` to verify differential handling of stale resumes (`hadResume` dependent) vs context overflow / poisoned transcript (`hadResume` independent) vs network errors.
  - **Behavioral fresh retry**: Asserts `runEngineTurn` retries once fresh with `resumeSessionId: null` upon encountering a stale resume target with no usage, successfully resetting the runner session and returning the clean attempt.
  - **No retry on usage**: Asserts `runEngineTurn` calls the engine **strictly once** and does not retry when `usage` is reported (e.g. mid-turn context overflow), preserving the turn error and usage.
  - **Edge-case safety**: Asserts errors with usage (even if quoting stale phrasing) and non-stale errors are not retried.

### 3. Rebase & CI
- Rebased on latest `main` (`v0.16.2`, commit `92ef7fa`).
- Verified locally:
  - `npm run lint` (Biome): clean, 516 files checked.
  - `npm run typecheck` and `npm run server:typecheck`: 0 errors.
  - Guards (`guard:big-brain`, `guard:llm-tracked`, `guard:engine-registry`): all passed.
  - All 284 computer unit tests passed.

wg2038:
Updated to address the empirical measurements and feedback from @WhichPaths on #209:

1. **Claude error extraction (`ClaudeSession.onStdout`)**:
   - Now checks `ev.errors` array in addition to `ev.result` and `ev.error`.
   - Expanded the fallback tail window from 5 lines to 30 lines (`.slice(-30)`), matching `failurePreview` so trailing output cannot push the cause out of view.

2. **Non-zero usage check (`hasNonZeroUsage`)**:
   - Claude 2.1.207 emits `usage: { input_tokens: 0, output_tokens: 0 }` on rejected resume targets. A naive `!result.usage` was truthy and prevented the safe retry from firing.
   - Replaced with `!hasNonZeroUsage(result.usage)`: turns with zero tokens consumed are safely retried once fresh; turns that consumed tokens (> 0) are strictly never retried in-turn to protect side effects and avoid double-billing.

3. **Belt-and-braces reset restored**:
   - Restored `!!resumeSessionId` at both `runAgenda` and `runTurn` post-turn reset points (`mustResetSession(engineError, !!resumeSessionId)`), ensuring stale resume targets that fail to recover in-turn are wiped from disk before the next wake.

4. **Regression test suite**:
   - Added dedicated tests covering `hasNonZeroUsage` token distinctions.
   - Added a regression test replaying the exact Claude 2.1.207 error shape (`errors` array, zero usage, stale resume) verifying in-turn fresh retry and disk session reset.

All checks, linters, and test suites are green.

yetone:
Closing this one, and I want to be clear that it is not because the work was wrong — your diagnosis in #209 was correct and it is what got this fixed.

#231 landed as a702da0 for the same issue. It arrived about 90 minutes after yours and ended up covering a strictly larger surface, including the one thing I asked you to change above:

- the fresh retry there is gated on a structured `failure.kind === 'resume-not-found'` only, so context overflow and poisoned transcripts keep their old reset-and-recover-next-wake path and a turn that already ran tools is never replayed;
- the `persistentUnusable` transport backstop is preserved with its `!result.usage` guard intact;
- session files are written atomically at `0600` and the legacy unscoped file is *quarantined* (renamed aside) rather than `rm`'d, so a support case can still recover it;
- there are tests that actually drive the retry state machine, including "ambiguous failures are never replayed" and "no third attempt".

Your two other findings are in that fix as well, because they were the real root cause and you identified both:

1. `ClaudeSession.onStdout` collapsing the error to `'see log'` and hiding `No conversation found with session ID` from the stale-session detector;
2. a stale resume wrongly setting `persistentUnusable` and wedging the agent across every later wake.

Thank you for the diagnosis — it was the hard part. #123 is still open and I would rather have your time there.

Issue #225, opened 2026-09-06
Request add Base URL for each agent


scaactk:
Currently, it can only use the default local Codex or Claude settings with the official base URL. Is it possible to add a field for entering a custom base URL?

WhichPaths:
A custom base URL does work today — just not per agent, which I think is the actual gap this is asking about. Worth separating the two, because the first half needs docs and the second needs a design decision.

**Claude.** `ANTHROPIC_BASE_URL` reaches the engine either from the daemon's own environment or from the `env` block of `~/.claude/settings.json`. Secure mode deliberately recovers that narrow set — `claude-user-settings.ts:17-25`:

```ts
export const CLAUDE_CORE_ENV_KEYS = [
  'ANTHROPIC_API_KEY',
  'ANTHROPIC_AUTH_TOKEN',
  'ANTHROPIC_BASE_URL',
  'ANTHROPIC_SMALL_FAST_MODEL',
  'ANTHROPIC_DEFAULT_OPUS_MODEL', 'ANTHROPIC_DEFAULT_SONNET_MODEL', 'ANTHROPIC_DEFAULT_HAIKU_MODEL',
] as const
```

with the file's own header explaining why: `--restricted` ignores user settings wholesale, "but custom providers commonly keep the small provider/bootstrap surface the Claude core needs in ~/.claude/settings.json."

The model picker knows about it too. A base URL that is not Anthropic's own sets `prefersLocalDefault`, which drops the opus/sonnet/haiku presets rather than claiming those aliases exist on someone else's endpoint, and `supportsCustom` lets you type the provider's own model id.

**Codex.** The core process gets the daemon's environment unmodified (`env: args.env`), so it reads your `~/.codex/config.toml` as usual, and the secure-mode `-c` overrides only set filesystem, permissions, `web_search` and `features.hooks` — none of them touch `model_provider`. `shell_environment_policy.inherit="none"` looks like it would block this but does not: that governs the environment handed to **model-spawned commands**, and the comment right above it says so — "The Codex core keeps its auth environment, while every model-spawned command receives only this explicit, non-secret environment."

I read this rather than ran it, so treat the Codex half as "should work" until someone confirms against a real custom provider — the Claude half I have exercised while working on #194.

**What is genuinely missing is scope.** Both routes are per-machine: one base URL for every agent on that computer. If you want Atlas on your own gateway and Iris on the official endpoint, on the same laptop, there is no way to express it today.

[...]

PR #233, opened 2026-09-07
feat(byoa): add per-agent Claude provider profiles

lyly-bonbon:
This is a first pass at #225, starting with Claude.

This adds a provider choice to the Agent editor, so Claude agents on the same computer can use different gateways and accounts. Profiles live in `~/.cumora/providers.json`. Credentials and endpoint URLs stay on that computer; Cumora receives only profile IDs, labels and model defaults.

Main turns and small-brain triage use the same profile. Switching profiles stops the old runner and keeps native sessions separate. If a profile is missing or invalid, the Agent stops rather than quietly falling back to the computer's default account. There is also a `--doctor --provider <id>` option to check a profile locally.

This first pass covers Claude in secure BYOA mode. It includes migration `0008_agent_provider_profile`, invalidates old runtime tokens on a profile change, and keeps profile-bound Agents away from older daemons. Setup and rollout details are in `docs/PROVIDER_PROFILES.md`.

Tested with lint, both typechecks, all three architecture guards, frontend and CLI builds, and the editor in a browser. Unit tests: 1,229 passed, 4 skipped. Integration tests: 348 passed, 5 skipped, with one Redis subscription timeout in `CLI document change events are published only after commit`. The same failure reproduces on the untouched base (`92ef7fa`); that file also passed on a separate rerun.

The process and environment checks use a fake Claude CLI. I haven't tested this against a real provider account yet.


lyly-bonbon:
Rebased onto main — this now sits on top of #235, and CI is green again.

Two things worth flagging from the rebase:

**Migration renumbered.** #235 took `0007_engine_defaults`, so this one moved to `0008_agent_provider_profile` and the manifest tip / supported range went to 8. The checksum hashes only the SQL, so it is unchanged.

**Priority call between the two features.** #235 gives a Computer per-engine default models; this PR gives an Agent a provider profile. Where they overlap I made the profile win by skipping `engine_defaults` entirely, on the same reasoning the PR already applies to the deploy-level `CUMORA_DEFAULT_*_MODEL` pin: the profile's endpoint owns its model namespace, and an engine-wide model name has no meaning inside it. The AgentEditor hint is suppressed to match — otherwise `engineDefaultModel` would shadow the profile's own default in the "follow engine default" hint and the UI would promise a model the server never sends. Easy to invert if you'd rather `engine_defaults` apply underneath a profile; it's two lines in `listAgentsForComputer` plus the hint.

On #225 — @WhichPaths is right that the two halves are separable, and this is only the per-agent half. The "a custom base URL already works today, it just isn't per agent" documentation gap is not addressed here and probably wants its own PR.

yetone:
I reviewed this properly and I want to merge it — it just needs a rebase first.

**What I checked and liked.** The boundary is drawn in the right place and enforced on both sides: `providerProfileMetadata` is an explicit projection, `sanitizeProviderProfiles` repeats it on untrusted input, and the server only ever accepts a validated profile *id* (`readProviderProfileId` → 400). Credentials genuinely never leave the machine. `readProviderProfiles` opens with `O_NOFOLLOW`, `fstat`s the descriptor it actually opened, and enforces `0600` + owner + size — that is the correct order, not the TOCTOU-prone `stat`-then-`open`. The URL check rejects embedded userinfo, query and fragment and requires https outside loopback. Error messages deliberately carry no JSON values or credential material. `providerProfileEnv` refusing to run at all under `CUMORA_BYOA_ALLOW_UNSANDBOXED=1`, and blanking `ANTHROPIC_*` / `CLAUDE_CODE_USE_*` / `CLAUDE_CORE_ENV_KEYS` before setting its own, is the right shape.

Migration `0008` is well-formed: bounds bumped together, checksum exported, and the `DROP TRIGGER` is safe because `0004` is the only creator and always precedes it. Rotating `runtime_assignment_id` on a `provider_profile` change is the correct call — a profile switch really is a credential-boundary change and old runtime tokens should die with it.

**What you need to do.** #231 landed as a702da0 and restructured the same region of `daemon.ts` (per-engine `EngineSessionStore`, `runEngineAttempt`, `runWithSessionRecovery`). Your branch now conflicts there. Please rebase onto `main`; the conflict is confined to `daemon.ts`. Note that `sessionFile` is gone — session state is now owned by `EngineSessionStore`, constructed per engine — so "switching profiles keeps native sessions separate" wants to compose with that rather than with the old single-file logic. Worth deciding explicitly whether a profile change should also start a fresh engine session store or keep the profile fingerprint as its own dimension.

Two small things while you are in there:

1. The description still says "includes migration `0007_agent_provider_profile`"; the file is `0008` since #235 took 0007. Cosmetic, but the PR body is what I read first.
2. You wrote that you have not tested against a real provider account yet. I am comfortable merging on that basis because the feature is opt-in and inert for agents with no profile, but please say so again after the rebase if it is still true, so it goes into the release notes honestly.

Ping me when CI is green after the rebase and I will merge it.

lyly-bonbon:
Rebased onto `main` at 171d407 (v0.17.0), on top of #231. The conflict was confined to `daemon.ts` as you said.

**How the profile composes with `EngineSessionStore`.** I kept the profile fingerprint as its own dimension rather than starting a fresh store: `EngineSessionStore` takes an optional fourth `scope` argument and the runner passes `providerProfileFingerprint(provider)`, so the pointer lives at `<sessions>/<agent>/<engine>[.<fingerprint>].session`. Same engine-scoped directory, same legacy quarantine, `resumeSessionId()`'s cross-engine guard untouched. A profile change already rebuilds the runner via `configMatches`, so the new runner opens a different pointer and the old profile's session is never offered to the new endpoint — and switching back to an unchanged profile can resume where it left off. Both chat and agenda now go through `runWithSessionRecovery`; the only thing I carried over from my side there is the `teardown.signal.throwIfAborted()` check before the attempt.

`ensureEngineSession` keeps your `resumeSessionId` parameter plus the `stopped` guard from this PR. The provider-profiles test was updated for the new pointer layout.

PR description now says `0008_agent_provider_profile`.

**Testing status, restated for the release notes:** still not exercised against a real provider account. The process and environment checks run against a fake Claude CLI; local typecheck, lint, guards and the computer/session test suites pass, and the feature is inert for agents with no profile.

yetone:
Rebase looks right and CI is green, so this is merged.

Two things I checked on the rebase specifically, since they were the parts I flagged:

**`EngineSessionStore` composition.** Keeping the profile fingerprint as a fourth `scope` dimension rather than swapping in a fresh store is the better of the two options I offered. `<sessions>/<agent>/<engine>[.<fingerprint>].session` means switching back to an unchanged profile resumes where it left off, which a fresh store would have thrown away, and the cross-engine guard in `resumeSessionId()` is untouched. `configMatches` already rebuilds the runner on a profile change, so the old profile's session id is never offered to the new endpoint — that is the property that matters and it falls out of the layout rather than being enforced separately.

**Migration 0008.** Bounds moved together, checksum regenerated, and `schema-migrations.test.ts` agrees. Note it lands in the same deploy as #123's 0009.

Merging with a merge commit rather than a squash, deliberately: #123 is stacked on this branch's commit, and squashing would have forced it into a fourth renumber. Not a precedent, just this pair.

Testing status recorded honestly for the release notes: not yet exercised against a real provider account; the process and environment checks run against a fake Claude CLI, and the feature is inert for agents with no profile. Thanks for the careful work on the credential boundary — the `O_NOFOLLOW` + `fstat`-the-descriptor order is the part most implementations get wrong.

PR #235, opened 2026-09-08
feat(byoa): per-engine default model configuration on computers

david-crazyamber:
## 背景 / Motivation

Users with custom provider endpoints (CC Switch and friends) need a per-engine model policy: which **main** + **fast** model each local CLI should run when an agent has no explicit pin. Custom endpoints often cannot enumerate their models, so the daemon's model catalog is empty and there is no durable place to name the models that actually work against the custom endpoint.

## 改动 / Changes

- **db**: versioned migration `0007_engine_defaults` adds `computers.engine_defaults` (JSONB map per engine: `{ model, fastModel }`)
- **server**: `listAgentsForComputer` model resolution priority becomes
  explicit agent pin → computer `engine_defaults` → reported local default (custom Claude endpoint) → deploy-level `CUMORA_DEFAULT_*_MODEL` pin
- **api**: `GET/PUT /computers/:id/engine-defaults` (merge-on-write; `null`/empty clears a field)
- **ui**: Me → Computers card gains a per-engine **"Configure models"** entry (main + fast inputs); AgentEditor "follow engine default" hint shows the configured engine default
- **daemon**: no change needed — the 60s roster poll picks up the resolved model and `sync()` rebuilds the runner on config change
- **docs**: `docs/BYOA.md` model-resolution section updated

## 验证 / Verification

- server + frontend `tsc` clean, `vite build` OK
- schema-migrations tests 11/11 (incl. new 0007 checksum), engine/computer unit tests pass
- rebased onto current `upstream/main`

## Note

Rebased on top of upstream main (v0.16.x). Base was ~v0.14.2, so migration number moved to `0007` (upstream already owns 0005/0006).

### DIFF
--- a/server/src/agents/computer/registry.ts
+++ b/server/src/agents/computer/registry.ts
@@ -1113,36 +1113,50 @@ export async function updateEngineDefaults(args: {
   companyId: string
   defaults: EngineDefaultsMap
 }): Promise<EngineDefaultsMap | null> {
-  const { rows } = await pool.query<{ engine_defaults: unknown }>(
-    `SELECT engine_defaults FROM computers
-      WHERE id = $1 AND company_id = $2 AND kind <> 'cloud' AND revoked_at IS NULL LIMIT 1`,
-    [args.computerId, args.companyId],
-  )
-  if (!rows[0]) return null
-  const existing = sanitizeEngineDefaults(rows[0].engine_defaults)
-  const merged = sanitizeEngineDefaults(args.defaults)
-  // Deep merge: for each engine, merge model and fastModel
-  for (const [engineId, defaults] of Object.entries(merged)) {
-    if (defaults.model === null && defaults.fastModel === null) {
-      // Both null means remove this engine's defaults entirely
-      delete existing[engineId]
-    } else {
-      existing[engineId] = {
-        ...(existing[engineId] ?? {}),
-        ...(defaults.model !== undefined ? { model: defaults.model } : {}),
-        ...(defaults.fastModel !== undefined ? { fastModel: defaults.fastModel } : {}),
-      }
-      // Clean up nulls
-      if (existing[engineId].model === null) delete existing[engineId].model
-      if (existing[engineId].fastModel === null) delete existing[engineId].fastModel
-      if (!existing[engineId].model && !existing[engineId].fastModel) delete existing[engineId]
+  const client = await pool.connect()
+  try {
+    await client.query('BEGIN')
+    // Lock before reading so concurrent partial updates merge into the latest state.
+    const { rows } = await client.query<{ engine_defaults: unknown }>(
+      `SELECT engine_defaults FROM computers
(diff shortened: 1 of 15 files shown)


server/src/agents/computer/registry.ts
```diff
@@ -1057,6 +1077,88 @@ export async function reportDetectedEngines(args: {
...
+        ...(defaults.fastModel !== undefined ? { fastModel: defaults.fastModel } : {}),
+      }
+      // Clean up nulls
+      if (existing[engineId].model === null) delete existing[engineId].model
+      if (existing[engineId].fastModel === null) delete existing[engineId].fastModel
+      if (!existing[engineId].model && !existing[engineId].fastModel) delete existing[engineId]
+    }
+  }
+  await pool.query(
+    `UPDATE computers SET engine_defaults = $2::jsonb WHERE id = $1`,
```
bingqilinweimaotai: Concurrent requests can read the same defaults and overwrite each other. If two admins save Claude and Codex settings at the same time, both requests succeed but the last write discards the other update.

src/components/AgentEditor.tsx
```diff
@@ -97,11 +97,16 @@ export function AgentEditor({ agent, onClose, onSaved }: Props) {
...
+  // When the local CLI catalog can't name a default (e.g. custom endpoints),
+  // these Cumora-side settings show as the hint for "follow engine default".
+  const engineDefaultModel = selectedComputer?.engineDefaults?.[selectedEngineId]?.model ?? undefined
+  const engineDefaultFastModel = selectedComputer?.engineDefaults?.[selectedEngineId]?.fastModel ?? undefined
   const modelOptions: Array<ComboboxOption<string>> = [
     {
       value: '',
       label: t('agent.followEngineDefault'),
-      hint: modelCatalog?.defaultModel ?? undefined,
+      hint: modelCatalog?.defaultModel ?? engineDefaultModel,
```
bingqilinweimaotai: The hint prefers the CLI default, but the server prefers the computer-level default. When both are set, the UI shows a different model from the one actually used. The fast-model hint below has the same issue.

src/locales/zh-CN.ts
```diff
@@ -1331,6 +1332,13 @@ export const zhCN: Partial<Record<keyof typeof en, string>> = {
   'me.computerStatus.online': '在线',
   'me.computerStatus.busy': '忙碌',
   'me.computerStatus.offline': '离线',
+  'me.engineModelConfig': '{engine} 的默认模型',
+  'me.engineModelNotSet': '跟随引擎默认',
+  'me.engineModelConfigure': '配置模型',
+  'me.engineDefaultModel': '主模型（大模型）— 留空则跟随 CLI 默认',
```
bingqilinweimaotai: Leaving this blank only clears the computer-level override. With an unpinned agent and a deployment default configured, the server can still select CUMORA_DEFAULT_*_MODEL instead of the CLI default. The English label makes the same promise.

PR #255, opened 2026-09-09
fix: unsubscribe ws listener in WorkspaceSettingsModal on unmount

AcWrong02:
ws.on() returns an unsubscribe function, but the useEffect did not return it as cleanup. Each open/close cycle leaked a listener that would trigger redundant loadMembers() calls on every membership event, accumulating over the session lifetime.

### DIFF
--- a/src/components/WorkspaceSettingsModal.tsx
+++ b/src/components/WorkspaceSettingsModal.tsx
@@ -52,10 +52,13 @@ export function WorkspaceSettingsModal({ company, companyCount, onInvite, onClos
     return () => { memberLoadGeneration.current += 1 }
   }, [loadMembers])
 
-  useEffect(() => ws.on((event) => {
-    if (event.type !== 'workspace.membership' || event.companyId !== company.id) return
-    void loadMembers(false)
-  }), [company.id, loadMembers])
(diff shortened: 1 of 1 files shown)


yetone:
Thanks for looking at listener lifetimes — that is a real class of bug and worth hunting. This particular one isn't there, though, and I'd rather explain why than just close it.

The original is a concise-body arrow function:

```ts
useEffect(() => ws.on((event) => {
  …
}), [company.id, loadMembers])
```

There are no braces around the effect body, so `ws.on(...)` is the arrow's *return value*, not a discarded statement. React receives the unsubscribe as the cleanup and calls it on unmount and before every re-run. Each open/close cycle already detaches its listener; nothing accumulates.

The shape is easy to misread — it looks like a call whose result goes nowhere, and the braced form you wrote makes the intent obvious at a glance. But the diff would land a commit message asserting a leak that never existed, and that message is what the next person reads when they `git blame` this line. So I'll leave it as is.

If you want to keep going on this: the pattern worth actually auditing is `useEffect(() => { ws.on(…) }, [])` — *braced*, result dropped. That one does leak, and it is the same two characters' difference. Worth a grep across `src/`.

PR #262, opened 2026-09-10
fix(documents): prevent data loss during snapshot compaction

bingqilinweimaotai:
Fix document compaction deleting persisted edits that are missing from the live room. Previously, the in-memory snapshot and the database MAX(id) could cover different updates, causing permanent data loss after a cold reload.

- Build snapshots from persisted Yjs updates read from a consistent database view.
- Serialize compactors with a document-level advisory lock and atomically replace the snapshot and delete only the merged log rows.
- Read the snapshot and all remaining updates in one statement during cold loading, including lower-ID updates that commit late.
- Trigger compaction after persistence succeeds and preserve the count of updates persisted during compaction.

Validation:
- Passed 7 PostgreSQL regression tests covering stale rooms, concurrent writes, out-of-order commits, concurrent compaction, rollback, and cold loading.
- Passed 15 document Markdown tests.
- Passed frontend and server type checks, repository-wide lint, and all three guard checks.

### DIFF
--- a/server/src/__integration__/document-compaction.test.ts
+++ b/server/src/__integration__/document-compaction.test.ts
@@ -1,0 +1,192 @@
+import assert from 'node:assert/strict'
+import { randomUUID } from 'node:crypto'
+import { after, before, beforeEach, test } from 'node:test'
+import { Pool, type PoolClient } from 'pg'
+import * as Y from 'yjs'
+import { compactDocument, readDocumentState } from '../documents/persistence.js'
+
+// Use an isolated schema: these tests need real PostgreSQL transaction and
+// sequence semantics but do not need the app's Redis or provider services.
+const schema = `compaction_test_${randomUUID().replaceAll('-', '')}`
+const connectionString = process.env.INTEGRATION_DATABASE_URL ?? process.env.DATABASE_URL
+const admin = new Pool({ connectionString })
+const pool = new Pool({ connectionString, options: `-c search_path=${schema}` })
+
+before(async () => {
+  await admin.query(`CREATE SCHEMA ${schema}`)
+  await pool.query(`
+    CREATE TABLE document_updates (
+      id BIGSERIAL PRIMARY KEY, document_id TEXT NOT NULL, update_bytes BYTEA NOT NULL
+    );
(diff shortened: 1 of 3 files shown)


yetone:
Merging. This is a real data-loss bug and the diagnosis is exactly right: `maybeCompact` snapshotted `Y.encodeStateAsUpdate(room.doc)` — the *live in-memory room* — but trimmed `document_updates` by `MAX(id)`. Those two cover different sets. Any update persisted by another instance that this room never applied was inside the `DELETE` range and outside the snapshot bytes, so it was gone for good on the next cold load. Building the snapshot from the persisted log instead closes it at the source rather than papering over it.

What I checked beyond the description:

- **The single-statement read is doing real work, not just tidiness.** `readDocumentState`'s `UNION ALL` puts the snapshot and the log on one MVCC view, so a compactor committing mid-read can't hand a cold load a snapshot from after the trim with a log from before it.
- **Reading *all* remaining updates, not `id > snapshot_at_update_id`, is the subtle half.** Sequence IDs are handed out before commit, so a transaction can commit late with an ID below the snapshot watermark. The old `loadUpdatesAfter` skipped exactly those. Good comment on it — that one will not be obvious to the next reader.
- **The delete is by explicit ID list, not a range.** A row inserted between the `SELECT` and the `DELETE` isn't in `updateIds`, so it survives. That's what makes the advisory lock sufficient rather than needing a stricter isolation level.
- **`updatesSinceSnapshot -= updatesAtStart` can't go negative** — the counter only ever increments elsewhere, and `updatesAtStart` is read at entry — so edits landing during a compaction are correctly carried into the next window instead of being zeroed.
- **The dropped `pool.connect()` in `hydrateDoc` is fine now.** That reservation existed because hydration was a two-query logical read and a pool-full wave could starve the tail query. It's one statement now, so the hazard it guarded is gone rather than ignored.

The tests are the reason I'm comfortable with this landing right before a release. Intercepting a query to hold a transaction open at a deterministic point, and then asserting on the *stale room* case, the late lower-ID commit, the between-read-and-delete commit, the losing compactor, and rollback — that's testing the interleavings that actually cause the loss, not just the happy path. All 9 checks green.

PR #263, opened 2026-09-10
fix(agent): the anti-monologue gate was eating the answer it asked for

WhichPaths:
## What a user sees

> I asked Nova in our #product room to draft the launch email. It replied **"Drafting the email now — ~30s"** and then went silent. The draft never arrived. Sometimes the room shows a red line instead: **"Agent run failed before it could finish. No result was produced."**
>
> The exact same request in a DM with Nova works fine.

## The product mandates a flow its own backstop rejects

`turn.ts` puts this in every turn's operating rules:

> **ANNOUNCE BEFORE LONG OR VISIBLE-EFFECT WORK** … POST A SHORT INTENT MESSAGE first via `cumora reply` … THEN do the work. THEN reply with the actual result. The intent message must be a **SEPARATE** `cumora reply` call, NOT chained with `&&`.

`cmdReply`'s anti-monologue gate then refuses exactly that second call:

```ts
const monologueBypass = Boolean(parsed.flags.continue || parsed.flags.also)
if (!monologueBypass && cv[0].actor_is_agent && cv[0].member_count > 2) {
  // …if my own message is last and younger than 10 minutes…
  return err(`you already posted in ${convoId} ${ageSec}s ago and nobody has replied yet …`)
```

Its exemptions are DMs, a ten-minute valve, and `--continue`. The announce-then-deliver flow is not among them — so in any room of three or more, following the rules costs the agent its answer.

## Where it turns destructive

The declared relay, because no model judgment is involved:

```ts
command: `cumora reply ${target.conversationId} ${escaped}`   // no flag
```

`relay.ok` is false → `finalStatus = 'failed'` → `postTurnFailureNotices` posts

> Agent run failed before it could finish (…). **No result was produced.**

into the room, in place of the work the agent had just done. The user asked, the agent said "on it", did the work, and the room got a red notice.

## The fix

Give the relay the bypass — `--continue`, appended last.

The gate exists to stop the agent **deciding** to speak again. Its own comment says why: *"each wake-up is a fresh 'should I respond?' decision with no global stop-signal."* The relay is not a decision. It is the runtime delivering text the model already composed and **explicitly declared** as this turn's reply (`assistant_text: "reply"` + `reply_conversation_id`), at most once per turn. That is precisely the deliberate commitment `--continue` is documented for:

> Bypass via `--continue` / `--also`: rare cases where the agent genuinely has to add to its own previous message … The flag forces the agent to commit deliberately rather than absent-mindedly continuing to monologue.

**The flag has to go last.** `parseArgs` reads `--continue <token>` as a *value* flag, so putting it before the body would consume the body and post an empty message. That is not hypothetical — it is what the parser does, and there is a test for it, because moving the flag to the front reads like a harmless tidy-up.

`flags.continue` is referenced in exactly one place in `cli.ts` (the gate), so the flag has no other effect.

## Scope, stated plainly

This fixes the deterministic path — the one where the answer is destroyed *and* replaced with a failure notice, with no model in the loop.

The model-driven path (the model calls `cumora reply` twice itself) still hits the gate and, per `personas.ts`, is told not to retry — so the answer is lost quietly rather than loudly. Reconciling that properly means either exempting a same-turn announce (which needs a run id the CLI does not currently have — messages carry no `run_id` and `executePodTool` does not plumb one) or relaxing the prompt rule. Both are your call, so I have left them alone rather than pick one on your behalf.

## Tests

Four drive the real `cmdReply` against Postgres, in `agent-anti-duplicate.test.ts` — the file that already owns this contract:

- **the gate refuses the result that the mandated intent message asked for** — the state being fixed, pinned so the interaction cannot silently return
- **the relay delivers that result, body intact** — asserts the stored body, not just `ok`
- **the bypass flag has to come after the body** — the `parseArgs` trap
- **a plain second post is still refused** — the guard rail: this must not turn the gate off

The fifth reads `turn.ts`, because the four above pass just as well against a relay that sends no flag. Verified red by removing the flag:

```
the declared relay no longer bypasses the anti-monologue gate — after the
intent message the rules require, the agent's answer is refused and the room
gets a failure notice instead
```

`agent-turn.test.ts`'s existing relay stub matched the command with a `$`-anchored regex. It now **requires** the flag rather than merely tolerating it, so that test becomes a second guard.

Green locally: `tsc --noEmit` (server + renderer), `biome lint .`, all three `scripts/guard-*.mjs`, the unit suite (1332 tests, 0 failures), the full `agent-anti-duplicate` suite (14/14), and all four declared-relay tests in `agent-turn`. The rest of `agent-turn` needs a real model endpoint — it is 17-of-24 red on clean `main` in my environment, so I am leaning on CI for that file rather than claiming a local result I do not have.

### DIFF
--- a/server/src/__integration__/agent-anti-duplicate.test.ts
+++ b/server/src/__integration__/agent-anti-duplicate.test.ts
@@ -264,3 +264,89 @@ test('[integration] worklog blocks a duplicate heavy-tool claim from a peer', as
     scopeKey, agentId: agentA, taskType: 'web-search', subject: 'audio editor competitive analysis',
   })
 })
+
+// ─── announce-then-deliver ─────────────────────────────────────────
+//
+// The operating rules in turn.ts REQUIRE an intent message before any work
+// that keeps the asker waiting: "POST A SHORT INTENT MESSAGE first via
+// `cumora reply` … THEN do the work. THEN reply with the actual result. The
+// intent message must be a SEPARATE `cumora reply` call."
+//
+// In a group of three or more, that intent message is then the room's last
+// message and seconds old — exactly what the anti-monologue gate refuses. So
+// the product mandated a flow its own backstop rejected, and the deliverable
+// that came back was the failure notice rather than the answer.
+
+test('[integration] the gate refuses the result that the mandated intent message asked for', async () => {
+  const { agentA, convoId } = await seedGroupWithTwoAgents()
+
+  const intent = await runCli(['--as', agentA, 'reply', convoId, 'Drafting the email now — ~30s'])
+  assert.equal(intent.ok, true, `the intent message the rules require must post: ${intent.text}`)
+
+  // …30 seconds of work later, the actual answer, exactly as the relay used to
+  // send it: no flag.
+  const answer = await runCli(['--as', agentA, 'reply', convoId, 'Here is the draft: ...'])
+  assert.equal(answer.ok, false, 'this is the state being fixed — the gate refuses the deliverable')
+  assert.match(answer.text, /you already posted in/)
+})
+
+test('[integration] the relay delivers that result, body intact', async () => {
+  const { agentA, convoId } = await seedGroupWithTwoAgents()
+  await runCli(['--as', agentA, 'reply', convoId, 'Drafting the email now — ~30s'])
+
+  // The shape turn.ts's declared relay now sends: flag LAST.
+  const body = 'Here is the draft: subject line, three paragraphs, CTA.'
+  const relayed = await runCli(['--as', agentA, 'reply', convoId, body, '--continue'])
+  assert.equal(relayed.ok, true, `the relay must deliver the answer: ${relayed.text}`)
+
+  const { rows } = await pool.query<{ body: string }>(
+    `SELECT body FROM messages WHERE conversation_id = $1 ORDER BY sequence DESC LIMIT 1`,
+    [convoId],
+  )
+  assert.equal(rows[0].body, body, 'the answer must arrive whole, not empty')
+})
+
+test('[integration] the bypass flag has to come after the body', async () => {
+  // parseArgs reads `--continue <token>` as a VALUE flag, so putting the flag
+  // before the body consumes the body: the gate is bypassed and an EMPTY
+  // message is posted. Pin the ordering, because moving the flag to the front
+  // reads like a harmless tidy-up.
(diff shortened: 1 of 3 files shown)


yetone:
Merging — the fix is right and the reasoning about *why* the relay deserves the bypass is the good part: the gate exists to stop the agent **deciding** to speak again, and the relay is the runtime delivering text the model already committed to. Also glad you checked that `--continue` doesn't weaken the verbatim-dup check; that one is deliberately not gated on `monologueBypass` (cli.ts:2312), so the guard rail holds.

**One correction to the repro, because it changes what's still broken.**

The description has the intent message and the relayed answer in the same turn. That path can't reach the relay. `postedReplyViaTool` is set the moment the intent `cumora reply` reports its side effect (turn.ts:3005), it is never reset within a turn, and the relay branch is guarded on `!postedReplyViaTool` (turn.ts:3212). So once the mandated intent message lands, the auto-relay is switched off for the rest of that turn — the second `cumora reply` is the model calling it directly, which is the model-driven path you've explicitly left out of scope.

The path that does produce the symptom you describe is **cross-wake-up**: the agent's own message from an *earlier* run is the room's last message and under ten minutes old when this run's relay goes out. `finalStatus` flips to `failed`, and `postTurnFailureNotices` (turn.ts:1727) puts "No result was produced." in the room in place of the answer. Same destructive outcome, same fix, different trigger — and it's worth having right in the commit message, because the next person to read this line will be reasoning about when the relay fires.

The consequence: the flow the operating rules actually mandate — announce, then deliver, **"keep going in this SAME turn"** (turn.ts:2229) — is still refused after this lands. That's the model-driven path, and it fails quietly rather than loudly. So the title claims a bit more than the diff delivers. I'd rather land this now and track the same-turn case separately than hold a correct fix.

On your two options for it: plumbing a run id so the gate can exempt a same-turn announce looks like the honest fix, since it distinguishes "this agent is continuing work it announced" from "this agent woke up and decided to talk again" — which is the distinction the gate is actually reaching for. Relaxing the prompt rule would trade away the typing-indicator effect that stops peers colliding on shared resources, which reads like the more expensive loss.

Noted for the record: the fifth test asserts on `turn.ts` source text, and the flag-ordering test only asserts inside `if (wrongOrder.ok)` so it can pass vacuously. You were upfront that the first four pass against an unflagged relay, which is why the source-reading one exists — fair trade for now.

PR #272, opened 2026-09-11
fix(agent): compaction must not hoist an answered steer to the newest-input slot

WhichPaths:
## The bug

Stage-2 compaction partitioned history by item **type**: every non-tool item sitting after the first tool item went into a `tailNonTool` bucket that both rebuild paths re-appended *after* all surviving pairs.

```js
const newHistory = [
  ...stage1History.slice(0, leadingNonToolEnd),
  summaryMarker,
  ...survivingTailPairs,
  ...tailNonTool,        // <- everything non-tool, moved to the end
]
```

So compaction reordered, not just deleted. That matters more than it looks, because of two things that are true together:

- The only non-tool items that ever appear mid-history are `role:'user'` items the runtime injects itself — the mid-turn steer render, the status-required nudge, and the completion-rejection nudge.
- There is **no assistant message in `history`** for the model to see its own reply in. Assistant text is deliberately never added as a history item (synthesizing one with no `item_id` / `part_id` / `responseId` corrupts history shape — the comment in `turn.ts` says so and names the earlier bug).

Put together: the *only* in-context evidence that a steer was already handled is that the `function_call` / `function_call_output` pairs answering it sit **after** it. The tail-move inverted exactly that.

## Reproduced

Against the real module, on the history shape `turn.ts` builds:

```
before: [0] user  book the team sync for 4pm and send the agenda
        c1 c2 c3 pairs  (create 4pm, draft agenda, reply "booked 4pm")
        [7] user  [Mid-turn update — new message from Alice] actually make it 5pm
        c4 c5 pairs     (move to 5pm, reply "moved to 5pm")

after:  [0] user  book the team sync for 4pm …
        [1] user  [auto-compaction · 3 earlier pairs summarized]
        [2..5]    c4 c5          ← the answer, now BEFORE the question
        [6] user  [Mid-turn update] actually make it 5pm     ← LAST ITEM
```

The steer's own text is written as an imperative fresh arrival — *"Re-assess your current plan in light of these and continue"* — so sitting in the newest-input slot it reads as a live instruction. The agent moves the event a second time and posts a second reply. `turn.ts` assigns the compacted array back to `history`, so it re-asks on **every remaining hop**, not once.

Anything side-effecting the steer asked for — send an email, post to a channel, move a calendar event — gets done twice. The same shape applies to the status-required and completion-rejection nudges: a satisfied nudge re-presented last is a live instruction again.

**How often:** it needs a mid-turn steer, at least one more hop of work after it, and stage-2 compaction firing (past 75% of the context window, with stage-1 truncation not enough on its own). Steering is a headline feature and long tool-heavy turns are exactly the ones that cross 75%, so this is a recurring bug on busy agents rather than a corner case. It cannot happen when compaction stops at stage 1.

## The fix

Rebuild in original order. `partitionHistory` no longer collects a tail bucket — it exists only to decide *which* `call_id` groups get dropped. Both rebuild sites now walk the original array and skip the dropped groups:

```ts
function tailWithoutDroppedGroups(history, leadingNonToolEnd, droppedCallIds) {
  const kept = []
  for (let i = leadingNonToolEnd; i < history.length; i++) {
    const item = history[i]
    const key = groupKeyOf(item)
    if (key !== null && droppedCallIds.has(key)) continue
    kept.push(item)
  }
  return kept
}
```

Surviving items keep their positions and their relative order. Documented as **invariant 4** in the module header: compaction only deletes, never reorders.

This also stops compaction re-grouping interleaved items. One hop pushes `fc1, fc2, out1, out2`; that is the shape every *uncompacted* hop already puts on the wire, so handing it back unchanged is strictly closer to known-good than the `fc1,out1,fc2,out2` the bucket rebuild produced. The existing invariants still hold — a `function_call` still immediately precedes its own `function_call_output`, and the leading seed messages are untouched.

## Tests

Three added, **no existing test changed** — 100 lines added, 0 removed:

- `stage 2 (drop path): an ANSWERED mid-turn steer keeps its position — never hoisted to the newest-input slot`
- `stage 2 (summary path): an ANSWERED mid-turn steer keeps its position`
- `stage 2: surviving tool items keep their original relative order (no regrouping by call_id)`

```
against main:   not ok 41, not ok 42, not ok 43        # pass 40  # fail 3
with the fix:                                          # pass 43  # fail 0
```

The 40 that were already there stay green, including the one that pins how a malformed no-`call_id` item is handled — that item's original position happens to be after the surviving pairs, so order-preserving and tail-moving agree there.

Also green: `tsc --noEmit`, `biome lint .`, all three `scripts/guard-*.mjs`, and the unit suite (1399 tests, 0 failures).

### DIFF
--- a/server/src/__tests__/agents-turn-compaction.test.ts
+++ b/server/src/__tests__/agents-turn-compaction.test.ts
@@ -776,3 +776,103 @@ test('summary: items with no call_id (legacy / bare function_call) do NOT confus
   assert.equal(last.type, 'function_call')
   assert.equal(last.name, 'orphan')
 })
+
+// ── Position preservation: an answered steer must not be re-asked ───────────
+//
+// Mid-turn steers (turn.ts `tryDrainSteer`) push the new user message into
+// `history` as a role:'user' item, and the tool calls that ANSWER it land
+// after it. That ordering is the only in-context evidence the steer was
+// handled: assistant text is deliberately never added as a history item
+// (turn.ts, "would corrupt history shape — see earlier bug"), so there is no
+// assistant message for the model to see its own reply in.
+//
+// Stage 2 used to bucket every non-tool item sitting after the first tool item
+// and re-append that bucket AFTER all surviving pairs. A steer answered in hop
+// 5 therefore reappeared as the LAST item of the hop-20 input — the fresh-input
+// slot — with its answer now ordered before the question, while its text reads
+// as an imperative fresh arrival ("Re-assess your current plan … respond to the
+// new message in its own conversation"). The agent replied to the user twice.
+
+const steerMsg = (text: string): ResponseInputItem => ({
+  type: 'message',
+  role: 'user',
+  content: text,
+}) as unknown as ResponseInputItem
+
+/** History shaped like a real steered turn: seed question, two old tool
+ *  pairs, the steer, the pair that answered the steer, one later pair. */
+const steeredHistory = (): ResponseInputItem[] => [
+  userMsg('original question'),
+  fnCall('call_old1'), fnOutput('call_old1', 'a'),
+  fnCall('call_old2'), fnOutput('call_old2', 'b'),
+  steerMsg('[Mid-turn update — 1 new message received while you were working]\n@ana (in conversation c9):\nalso check the logs\nRe-assess your current plan in light of these and continue.'),
+  fnCall('call_answer'), fnOutput('call_answer', 'posted reply to c9'),
+  fnCall('call_later'), fnOutput('call_later', 'd'),
+]
+
+const indexOfSteer = (history: ResponseInputItem[]): number =>
+  history.findIndex((it) => {
+    const r = it as unknown as { content?: unknown }
+    return typeof r.content === 'string' && r.content.startsWith('[Mid-turn update')
+  })
+
+const indexOfCallId = (history: ResponseInputItem[], callId: string): number =>
+  history.findIndex((it) => (it as unknown as { call_id?: string }).call_id === callId)
+
+test('stage 2 (drop path): an ANSWERED mid-turn steer keeps its position — never hoisted to the newest-input slot', () => {
+  const result = compactHistory(steeredHistory(), ALWAYS_OVER)
+  assert.equal(result.droppedPairCount, 2, 'the two oldest pairs go; the steer and its answer survive')
+
+  const steerAt = indexOfSteer(result.newHistory)
+  assert.ok(steerAt >= 0, 'the steer itself must never be dropped')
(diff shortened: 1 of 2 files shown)


yetone:
Verified and merging.

I reproduced the bug from the source rather than taking the description on trust. Both rebuild sites did append the bucket last — `turn-compaction.ts:318-322` (drop path) and `:362-366` (summary path) — so a mid-history item really was relocated to the final slot.

The premise that makes it harmful checks out too. There are exactly three mid-history non-tool injection sites in `turn.ts`, all `role:user`, and all three are imperatives rather than statements:

- `:2369` the steer render — *"Re-assess your current plan in light of these and continue. CRITICAL — ... (a) respond to the new message in its own conversation"*
- `:2804` the status-required nudge — *"Continue this same turn and call set_turn_status"*
- `:3127` the completion-rejection nudge — *"Continue this same turn."*

Hoisted into the newest-input slot, each of those reads as a live instruction, and the only counter-evidence — its answering pairs sitting after it — is exactly what the move destroyed. Your test fixture is a faithful model of the `:2369` shape.

I confirmed the tests are not tautological by running your test file against unmodified `main`:

```
not ok 41 - stage 2 (drop path): an ANSWERED mid-turn steer keeps its position …
  error: the answered steer was hoisted into the newest-input slot …
not ok 42 - stage 2 (summary path): an ANSWERED mid-turn steer keeps its position
not ok 43 - stage 2: surviving tool items keep their original relative order
# pass 40  # fail 3
```

and then with `turn-compaction.ts` from this branch: `# pass 43  # fail 0`. The 40 pre-existing tests pass unchanged in both.

Two things I want to credit, because they are the parts that would have bitten a careless version of this fix:

1. Pinning `stage1History` before the loop. The drop path reassigns `newHistory` each iteration, so rebuilding from it would have compounded across iterations once the marker moved into the tail source. The comment says exactly why.
2. Dropping the `pairItems.delete(cid)` mutation in favour of a `droppedCallIds` set. `pairItems` is now read-only bookkeeping for `droppedItemBytes`, and since each `cid` in `dropEligible` is distinct there is no double counting.

Invariant 1 still holds by construction, as your doc comment argues: membership is decided per `call_id` group, so an output can never survive without its call.

One nit, not worth a change: `steerMsg` builds `content` as a bare string, while `turn.ts:2369` pushes `content: [{ type: 'input_text', text }]`. `groupKeyOf` only inspects `call_id`, so both are equally non-tool and the fix is shape-agnostic — but the fixture is one step away from the production shape, and `indexOfSteer`'s `typeof r.content === 'string'` bakes that in. If a future test needs to assert on steer *content*, it will want the array form.

PR #275, opened 2026-09-15
fix: backfill missing message history after reconnect

bingqilinweimaotai:
## Summary

Reconnecting previously merged only the latest 80 messages with cached history. When more than one page arrived while disconnected, this left gaps that older-message pagination could never reach. For example, cached messages 1–80 became 1–80 + 121–200, leaving 81–120 inaccessible.

Fetch additional pages back to the latest cached message before merging. Apply this to both reconnect reloads and loading invalidated conversations when switching back to them.

## Tests

Add 16 regression tests in `tests/messages-history.test.ts`, covering:
- Both loading paths, page boundaries, and multi-page gaps.
- Empty caches and optimistic messages.
- Failed backfill followed by retry.
- Messages arriving during backfill.
- Older-history pagination and scroll anchoring.

Verified that the regression tests detect the original bug.

## Validation

- `node --import tsx --test "tests/**/*.test.ts"` — 77 tests passed.
- `npm run typecheck` — passed.
- `git diff --check` — passed.


yetone (approved):
Reviewed and verified locally on top of current `main` (`a030961`):

- `tsc --noEmit` clean; full unit suite 1413 pass / 0 fail with the 16 new tests included.
- Read `fetchMessagesSinceCache` against the server contract in `router.ts` (`GET /messages` returns ASC, `before=<seq>` is strictly older) — the loop's `page[0].sequence > latest` cursor is correct for that ordering, and filtering to `> latest` keeps the older-history/scroll-anchor responsibility with `loadOlder`, which the last test pins down.
- The optimistic-`MAX_SAFE_INTEGER` exclusion and the "cache untouched on failed backfill" behaviour are exactly what I'd want; thank you for making the failure test non-tautological.

One thing to keep in mind rather than a blocker: the backfill is unbounded — a desktop client that sat open through a very busy room and then reconnects will page sequentially all the way back to its cached tail before the merge lands. That is strictly better than the unreachable gap it replaces, and each page is a cheap indexed read, so I'm landing it as-is. If it ever shows up as a slow reconnect in practice, a page cap that falls back to dropping the cache is the follow-up.

Merging — this goes out in v0.18.3.

PR #278, opened 2026-09-15
fix(docs): doc create --body never reached the database

WhichPaths:
## The bug

`cumora doc create --body` loses the body about half the time, and the CLI reports success every time.

`cmdDoc` runs the whole command in one transaction. It inserts the `documents` row on that client, then seeds the body:

```ts
await dbClient.query(`INSERT INTO documents (id, company_id, title, created_by) VALUES ($1,$2,$3,$4)`, …)
const body = typeof parsed.flags.body === 'string' ? unescapeChat(parsed.flags.body) : ''
if (body) {
  const { applyAgentEdit } = await import('../documents/rooms.js')
  await applyAgentEdit(id, companyId, me, [{ kind: 'append', text: body }], dbClient)
}
```

The room's Yjs update hook persists that seed with the **global pool** — a different connection, which cannot see the still-uncommitted parent row:

```ts
async function persistUpdate(documentId, authorId, bytes) {
  await pool.query(`INSERT INTO document_updates (document_id, author_id, update_bytes) VALUES ($1,$2,$3)`, …)
}
…
void persistUpdate(documentId, authorId, update).catch((e) => {
  console.warn('[docs] persistUpdate failed', e)      // ← swallowed
})
```

Postgres does not wait for the parent either. There is no row to lock, so the FK check fails immediately — verified against the project's own PG 16 container:

```
BEGIN; INSERT INTO fk_parent VALUES ('p1');  -- held open, uncommitted
-- on a different connection:
INSERT INTO fk_child (parent) VALUES ('p1');
ERROR:  insert or update on table "fk_child" violates foreign key constraint
elapsed: 0s   (it failed immediately; it did not block)
```

The command still returns `ok` with `bodyLength` set.

## Why nobody noticed

The in-memory room still holds the body, and the agent paths (`applyAgentEdit` / `readDocumentText` / `collectDocumentStorageKeys`) never subscribe — the grace-window eviction in `unsubscribe` is the only thing that schedules eviction. So `doc read` and every browser on that instance keep serving the body from RAM until the process dies.

## It is not just the body

The seed is the parent struct for everything appended afterwards, so later edits persist as rows that can never integrate. Measured on a document whose seed was lost:

```
victim (create lost) = doc_5fdbd5eb3dd34559
doc append           → ok = true,  document_updates rows = 1   ← the row IS written
live-room text       = "BODY-0\nHUMAN-AND-AGENT-WORK-AFTER"
COLD text            = ""                                      ← everything gone
```

So a team can work in that document for days, and the next restart empties it.

`--body` is the documented way for an agent to create a doc with content — the help text, the operating rules in `turn.ts`, and `personas.ts` all instruct `cumora doc create '<title>' --body '<markdown>'`.

## The fix

When `applyAgentEdit` is given a transaction client, it takes the update out of the room's fire-and-forget persist path and writes it on that client itself — awaited, inside the caller's transaction, and not swallowed:

```ts
const collected: Uint8Array[] = []
const origin = (dbClient
  ? { originId: `agent:${agentId}`, authorId: agentId, collectInto: collected }
  : { originId: `agent:${agentId}`, authorId: agentId }) as never
…
for (const update of collected) {
  if (!dbClient) break
  await persistUpdateWith(dbClient, documentId, agentId, update)
  room.updatesSinceSnapshot += 1
}
```

The FK passes because the transaction sees its own parent row; the seed is atomic with the document row; and a failed write now fails the command instead of reporting success over a body that never landed.

Edits with **no** client — every browser edit, and any agent edit outside a transaction — keep the existing pool path unchanged. Fan-out to other instances is unchanged for both paths.

**Measured before: 11/20 creates lost the body. After: 0/20.**

## Tests

New `server/src/__integration__/document-agent-create.test.ts`:

- `doc create --body persists the body, every time`
- `the body is still there after a restart` — evict the room, re-hydrate from the DB
- `work appended later survives too` — the severity multiplier
- `a doc created without --body still works` — the guard rail: the ordinary create produces no seed update at all

They loop ten times on purpose. The defect is a race between the seed's `INSERT` and the command's `COMMIT`, so one run reproduces it only about half the time; ten make both directions decisive:

```
against main:   not ok 1 - doc create --body persists the body, every time
                  10/10 bodies were reported as created and never written to document_updates
                not ok 2 - the body is still there after a restart
                not ok 3 - work appended later survives too
                ok 4     - a doc created without --body still works
                # pass 1  # fail 3

with the fix:   # pass 4  # fail 0
```

Also green: `tsc --noEmit` for server and renderer, `biome lint .`, all three `scripts/guard-*.mjs`, the unit suite (1417 tests, 0 failures), and the neighbouring document suites — `document-compaction`, `document-image-cleanup`, `ws-doc-fanout`, `ws-doc-authorization` (18/18).


yetone (approved):
Reproduced and verified — approving.

**Premises checked:** `cmdDoc` does run the `INSERT INTO documents` and `applyAgentEdit` on the same transaction client (`cli.ts:6178-6188`); `document_updates.document_id` is a real FK to `documents(id)` (`migrate.ts:1491`); and the old persist path went out on `pool`, so the FK check ran on a connection that could not see the uncommitted parent. Postgres does not block on an uncommitted parent for an FK check, so the "fails immediately, swallowed by `console.warn`" story holds.

**Mechanism checked:** Yjs fires `update` synchronously inside `doc.transact`, so `collected` is fully populated before the awaited `persistUpdateWith` loop runs; browser edits and out-of-transaction agent edits (no `dbClient`) keep the pool path byte-for-byte; fan-out is untouched on both branches.

**Locally, against the project's PG 16 (`cumora_test`):**
- `document-agent-create.test.ts` on `main`'s `rooms.ts`: `not ok 1/2/3` — "8/10 bodies were reported as created and never written to document_updates"
- with this branch: `# pass 4 # fail 0`
- all five document integration suites together: `# pass 22 # fail 0`
- `biome lint`, `server:typecheck` clean

One non-blocking note: the `maybeCompact` kick after the tx-client persist runs on the pool and therefore cannot see the row it was told about until the caller commits. That is harmless — a snapshot is built from the in-memory doc, and a Yjs update replayed over a snapshot that already contains it is a no-op — but it does mean `updatesSinceSnapshot` can run a beat ahead of what the DB shows. Not worth holding this for.

Thanks — this was a real data-loss bug hiding behind a green CLI result.

PR #279, opened 2026-09-15
fix(messages): the backfill watermark has to mean complete, not biggest

WhichPaths:
Follow-on to #275, which added the reconnect backfill. The bridging loop is right; the watermark it bridges from is not.

## The bug

```ts
let latest: number | null = null
for (const message of current) {
  const seq = sequenceOf(message)
  if (seq !== null && seq !== Number.MAX_SAFE_INTEGER && (latest === null || seq > latest)) latest = seq
}
…
while (latest !== null && page.length === MESSAGES_PAGE_SIZE && page[0].sequence > latest) { … }
```

`latest` is the **biggest sequence the cache holds**. The loop needs **the newest sequence below which the cache is complete**. After an outage those are not the same thing, and one message is enough to separate them.

On reconnect the store does the right thing for the conversations the user is not looking at:

```ts
useMessages.setState((s) => ({ loaded: new Set(active && s.loaded.has(active) ? [active] : []) }))
if (active) void useMessages.getState().reloadConversation(active)
```

— `loaded` is invalidated, so the conversation backfills when the user next opens it. But `applyEvent` appends live `message.new` rows to `byConvo` for **any** cached conversation, loaded or not, and those land before the user gets there. That row becomes the max:

```
seen through 100 · offline · 101..200 arrive · reconnect · 201 arrives live
open it: newest page = 122..201 → page[0] = 122 → 122 > 201 is false
→ the loop never runs; 101..121 are missing for the rest of the session
```

Nothing else fills it. `loadConversation` short-circuits once loaded, and `refreshActiveMessagesIfSidebarMoved` only refetches when the sidebar's `lastMessageId` is absent from the cache — the live row is exactly what put it there.

**How often:** a conversation opened earlier, left for another one, an outage long enough to accumulate more than a page (80) of messages, and at least one message arriving after the reconnect before the user opens it. Laptop sleep/wake plus a busy agent channel gets there regularly.

## The fix

Walk the cache oldest-first and stop at the first discontinuity:

```ts
function completeThrough(current: Message[]): number | null {
  const sequences = current.map(sequenceOf)
    .filter((seq): seq is number => seq !== null && seq !== Number.MAX_SAFE_INTEGER)
    .sort((a, b) => a - b)
  if (sequences.length === 0) return null
  let through = sequences[0]
  for (const seq of sequences) {
    if (seq === through || seq === through + 1) through = seq
    else break
  }
  return through
}
```

That run is what the client can actually vouch for, and a message that arrived after the gap is the island it should be.

A rolled-back `INSERT` leaves a genuine hole and truncates the answer early. That costs one extra page and nothing else — the bridge now also filters by id, so rows already held are not re-added. Erring toward re-fetching is the safe direction; erring toward "caught up" loses messages.

**What does not work:** stopping on "this page contains an id we already hold". The live row is in the *newest* page, so the very first check would read the gap as already bridged. I tried that first.

## Tests

Two added to `tests/messages-history.test.ts`, using the harness #275 built:

- `backfills even when a live message arrived after the reconnect` — the defect. Red on main under **both** `reloadConversation` and `loadConversation`.
- `bridges across a hole in the cache without duplicating what is held` — the guard rail for the design decision above: ordered, no duplicates, gap still bridged. It passes either way on purpose.

**All 17 existing tests are unchanged and still pass**, including their exact expected backfill cursor sequences (`[undefined, 321, 241, 161, 81]` and friends) — the change moves the starting watermark, not the loop.

```
against main:  not ok 7 - backfills even when a live message arrived after the reconnect
                 the outage was never bridged because one live message raised the watermark past it
               (fails under reloadConversation and loadConversation)

with the fix:  # tests 19  # pass 19  # fail 0
```

Also green: `tsc --noEmit` for renderer and server, `biome lint .`, all three `scripts/guard-*.mjs`, and the unit suite (1420 tests, 0 failures).


yetone (approved):
Good catch, and the analysis is right: `loaded` is invalidated on reconnect but `applyEvent` still appends live rows to any cached conversation, so max(sequence) stops meaning "complete through". Approving.

**Checked the one thing that would make "stop at the first discontinuity" expensive:** whether sequences can have routine holes. They cannot in practice — the counter bump (`conversation_counters … RETURNING next_sequence - 1`) and the `INSERT INTO messages` are in the same transaction under `FOR UPDATE OF c`, so a rollback takes the counter with it; there is no hard delete of `messages`; and the HTTP list endpoint has no per-recipient filter (`WHERE m.conversation_id = $1`), so even `delivery_recipient_id` system rows are visible to everyone. So a hole is genuinely exceptional, and when one exists the re-walk is bounded by what the user had scrolled — the by-id filter plus `mergeFetchedMessages`'s id/clientId dedupe keep it duplicate-free. The "one extra page" line in the description is optimistic for a hole far back in a long cache, but it is the safe direction and the case is theoretical, so no change asked.

**Locally:** new tests red against `main`'s store (`# pass 17 # fail 2` — the new case under both `reloadConversation` and `loadConversation`), `# pass 19 # fail 0` with the fix; existing expected cursor sequences unchanged; `biome lint` and `tsc --noEmit` clean.

This lands in v0.18.4 immediately after v0.18.3, which shipped #275 minutes ago.

PR #294, opened 2026-09-19
feat(projects): permanently delete archived projects and scoped memories

bingqilinweimaotai:
Archived projects could previously only be restored. This adds a red Delete button and confirmation prompt so workspace owners and admins can permanently delete archived projects and their dedicated memories.

- Delete the project and its dedicated memories atomically, including pinned project memories.
- Preserve global memories, skills, and linked conversations; detach conversations from the deleted project.
- Clean up local project memories through computer heartbeats, with retries after offline devices reconnect.
- Prevent stale tasks from recreating database memories for deleted projects.
- Keep all memories when archiving a project.

Validation:
- Frontend and backend type checks passed.
- 42 related tests passed.
- Isolated SQL checks passed for scoped cleanup, tenant isolation, stale-write protection, and transaction rollback.

Deployment requires running `npm run migrate` to apply migration 0010 and updating paired computer clients to enable local memory cleanup.


yetone (changes requested):
Thanks for this — the daemon-side cleanup (`project-memory-cleanup.ts` with the component-by-component `lstat` + `realpath` containment checks) and the transactional `deleteArchivedProject` are careful work, and CI is green. I read the whole diff (migration + trigger, `project-deletion.ts`, router, daemon, `MeView`, tests). I'm **not merging it as-is**, for production-load reasons rather than correctness ones. Concretely:

### 1. Extra query on every heartbeat, for every computer, forever

`POST /computers/heartbeat` now runs `SELECT project_id, agent_ids FROM project_memory_deletions WHERE company_id = $1` on each tick. We have ~870 live computers heartbeating every 30 s against a Cloud SQL instance that was pegged at 99–100 % CPU until #290/#291 landed this week (now ~79 %, still not healthy). That's ~30 extra queries/s in steady state for an action that will happen a handful of times per month. And because rows in `project_memory_deletions` are never expired or acknowledged, the list only grows and is re-sent to every daemon on every heartbeat, including daemons that already cleaned up.

**Suggestion:** don't piggyback on the heartbeat. Either
- push the deletion to the affected daemons via the existing wake-stream (a `project.deleted` event, like other control events), with a bounded re-check on daemon *startup* only, or
- keep a pull model but have the daemon fetch `/computers/deleted-project-memories?since=<cursor>` on startup + on an explicit wake, and let the server drop rows after N days / after all agents acknowledged.

Either way the steady-state per-heartbeat cost must be zero.

### 2. Trigger with advisory locks on `agent_workspace` writes

`agent_workspace_deleted_project_guard` is a BEFORE INSERT OR UPDATE plpgsql trigger that takes `pg_advisory_xact_lock(...)` and does a lookup on every write to a `memory/` path. `agent_workspace` is one of the hottest write paths we have (every memory write from every agent, cloud + BYOA). Adding a trigger + transaction-scoped advisory lock there to close a race that only exists during the few seconds a project is being deleted is the wrong trade: it's a permanent tax on a hot path to guard a rare admin action, it's the first trigger-based invariant in the schema (all other invariants live in application code), and it's invisible to anyone reading the TS write path later.

**Suggestion:** drop the trigger and the advisory locks entirely. Do the check in the application memory-write path (one indexed `EXISTS` against `project_memory_deletions`, or simply re-check `projects` exists before the write), and accept that a write racing the delete may leave a row that the *next* delete/cleanup sweep removes. A rare orphaned memory row is much cheaper than a hot-path trigger.

### 3. Schema bump to `MIN=MAX=10`

The migration bumps `MIN_SUPPORTED_SCHEMA_VERSION` and `MAX_SUPPORTED_SCHEMA_VERSION` to 10/10. That means the currently running server (which supports 9/9) refuses to start against the migrated DB, so the deploy's automatic rollback path (migration Job → rollout → smoke → auto-restore) becomes a one-way door: if smoke fails, restoring the previous image fails schema verification. With the DB still under load and cold pods already flaky at boot, I don't want a one-way migration for a feature this size.

**Suggestion:** if a table is still needed after (1)/(2), ship it as `MIN=9, MAX=10` (additive; old code tolerates the new table), and only narrow MIN in a later release once 10 is proven in prod.

### 4. Smaller things (fine to fix alongside)

- `DELETE /projects/:id` deletes the project row itself, not only the memories. Since `conversations.project_id` / `shipping_features.project_id` are `ON DELETE SET NULL`, conversations silently lose their project association. The PR title says "delete archived projects", so this may be intended — but the UI `confirm()` text should say that conversations will be detached, and I'd rather see a soft `deleted_at` on `projects` than a hard delete.
- `deleteArchivedProject` writes **all** company agent ids into `agent_ids`, even agents that never had memory for that project; combined with (1) that fans out to every daemon in the company.
- Browser `confirm()` for a destructive, irreversible action is below the bar of the rest of `MeView` — use the existing dialog component with the project name typed back.

### What would make this mergeable

A version that (a) adds no per-heartbeat query, (b) has no DB trigger / advisory locks, (c) is either migration-free or a widening `MIN=9/MAX=10` migration, and (d) keeps the daemon cleanup code as-is. If you want to split it, the daemon cleanup + `deleteArchivedProject` (minus the trigger) + a wake-stream event is a good first PR; the schema/table can come later if it turns out to be needed.

Happy to re-review quickly once it's reworked. Leaving the PR open.

